PR #205, opened 2013-12-10
Throw exception and set approriate status if exception occurs in hook

### LINKED ISSUE #197: Hooks processing should be more exception safe
Currently it is up to the user in "environment.py" to ensure that no exceptions occur (by catching all). It would be better that hook errors (exceptions) would be caught and reported.

vrutkovs:
Fixes #197


jenisys:
Thanks.

With adaptions:
* Added hook_failed attribute to Feature, Scenario, Step (in behave model)
* Changed hook-error logic slightly to make more sense (at least for me ;-)

PR #219, opened 2014-04-15
IDE's unknown modules import issue

xbx:
For example, PyCharm can't detect which modules this packages own because their names are string (then splitted).
Putting them into a list, makes IDE more happier. 
Does it make sense?


jenisys:
INTEGRATED-INTO: behave-1.2.5a1

PR #281, opened 2015-01-16
Feature/background hooks

aidan:
Sometimes backgrounds get very expensive and it would be nice to be able to, for instance, run them once, dump the database afterwards (but before any further steps) and then reuse that for the remaining runs.

this PR adds before_background, after_background hooks for environment.py and a Background.skip()

### DIFF
--- a/behave/model.py
+++ b/behave/model.py
@@ -656,6 +656,51 @@ class Scenario(TagAndStatusStatement, Replayable):
             self.set_status(Status.skipped)
         assert self.status in self.final_status #< skipped, failed or passed
 
+    def _run_steps(self, steps, run_steps, runner, failed,
+                   dry_run_scenario):
+        for step in steps:
+            if run_steps:
+                if not step.run(runner):
+                    # -- CASE: Failed or undefined step
+                    #    Optionally continue_after_failed_step if enabled.
+                    #    But disable run_steps after undefined-step.
+                    run_steps = (self.continue_after_failed_step and
+                                 step.status == Status.failed)
+                    failed = True
+                    # pylint: disable=protected-access
+                    runner.context._set_root_attribute("failed", True)
+                    self.set_status(Status.failed)
+                elif self.should_skip:
+                    # -- CASE: Step skipped remaining scenario.
+                    # assert self.status == Status.skipped
+                    run_steps = False
+            elif failed or dry_run_scenario:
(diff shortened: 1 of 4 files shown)


behave/model.py
```diff
@@ -843,6 +848,38 @@ def skip(self, reason=None, require_not_executed=False):
             self._cached_status = "skipped"
         assert self.status in self.final_status #< skipped, failed or passed
 
+    # ASFIXME: badly name
```
frac: _run_main_steps  maybe?

aidan: Does both main and background steps, but not at the same time

frac: true

jenisys:
Note that your solution has a problem.
It leads to faulty summary computation when all steps passes and background steps are skipped from the second scenario (your third scenario with passing steps).

```!gherkin
1 scenario passed, 0 failed, 1 skipped
```

SHOULD BE:
```!gherkin
2 scenarios passed, 0 failed, 0 skipped
```
NOTES:
* Scenario status computation currently assumes that the first non-passed step.status leads to the scenario.status result.
* You don't need failing steps to see that the hooks are called. Use ``--no-capture`` option instead.
* I am also not sure if the skipped background steps should be shown at all
* You primary problem can already be solved by using the first scenario as "background once" substitute (with @setup tag). This will automatically include it when you use a file location. When you use tags, you need to include the setup tag manually

aidan:
I've fixed the status computation and will alter the test scenarios to use --no-capture. I'm also back and forth on showing the skipped steps.

I'm confused by the reference to @setup tag - is there an example of this pattern you could point me to?

iElectric:
+1 and some docs :)

AudriusButkevicius:
+1 
Needs updates in the runner operation in the docs.

aidan:
I've fixed those tests and amended context_attributes.rst - is there somewhere else this sould be mentioned?

aidan:
ping?

shyukri:
Any progress on that?

aidan:
@jenisys is this ok to merge now?

piotrweglinski-silvair:
Will this be merged ? Or it is decided that this will not be pulled into master?

j3ffx-hl:
Leaving a bottle at sea but, @jenisys any updates on why this?

Issue #280, opened 2015-01-13
AmbiguousStep error with similar step definitions and use_step_matcher("re")


tomekwszelaki:
I have two steps that are very similar - their definitions actually start with the same text, the difference being that the longer one has an additional world, like this:

```
use_step_matcher("re")

@when("I do stuff slowly")
def when_I_do_stuff_slowly(context):
    pass

@when("I do stuff")
def when_I_do_stuff(context):
    pass
```

If I define them as above the tests execute without any problem. However, if you switch them around you will get an AmbiguousStep error.

```
behave.step_registry.AmbiguousStep: @when('I do stuff slowly') has already been defined in
  existing step @when('I do stuff') at features\steps\uno.py:23
```

 It's because the step definition is taken "as is" and is used as a matcher here: https://github.com/behave/behave/blob/81aaea0d054aab5a9f2e6347b47d0a130fe1816f/behave/step_registry.py#L43

Do you think it would be possible to use the " ^ " and " $ "  delimiters to match strings exactly or is that impossible due to capturing arbitrary parameters from step definitions?

jenisys:
When you use the "re" step matcher, you can use the special begin/end delimiters in regular expressions directly, like:

```!python
from behave import given, when, then, use_step_matcher

use_step_matcher("re")

@step(u'^I do stuff$')
def step_impl(context):
    pass
```

NOTE: 
That does not work with the "parse" or the "cfparse" matcher because the expect the special chars literately. Therefore, only the ordering may help you,

tomekwszelaki:
Yeah, I tried that and it worked, but it's not an elegant solution since you only have to remember about it when you use a "re" matcher and only when you have a step definition that another step definition "contains" and only if the shorter one is defined before the longer one. 
So in 99% of cases you would write the step definition "as is" (meaning copy/paste directly from the .feature file) and then on a special occasion you would have to put the delimiters. I think it would be confusing.

Is there any way behave could do it internally so that you could just write step definitions exactly as they appear in the .feature file?

jenisys:
Should be rather simple. Instead of doing it yourself in the step definition (how its done in Cucumber), you move the functionality  into ``behave.matchers.RegexMatcher``. The ``behave.matchers.ParseMatcher`` does already do this inside of the ``parse`` module.

```!python
# -- WITH A EXTENDED RegexMatcher2.__init__(self, …):
#        expression = "^%s$" % self.string
#        self.regex = re.compile(expression)
[...]

PR #282, opened 2015-01-16
Add delimiters to regexp parser

tomekwszelaki:
Added the beginning and end string delimiters to the RegexMatcher. This way, the RegexMatcher will match whole strings and will not throw an AmbiguousStep erros when there are two step definitions beginning with the same string (see https://github.com/behave/behave/issues/280 for more details)


jenisys:
INTEGRATED INTO: behave-1.2.5a1

PR #283, opened 2015-01-16
fix fork me image in docs

frodopwns:
I noticed the "fork me" image was broken.  Seems like other users were using these:

https://github.com/blog/273-github-ribbons


jenisys:
INTEGRATED INTO: behave-1.2.5a1

Thanks.

PR #291, opened 2015-02-20
tiny quick wins while reading the tutorial

bernardpaulus:
Hello!

I started editing the docs Initially for a missing 'n' to given.
I then used this occasion to put some annotation mentions between code escapes.

Thanks for the very good doc btw :)


jenisys:
INTEGRATED INTO: behave-1.2.6.dev0 (with adaptations).

PR #292, opened 2015-02-21
Fix tutorial document mistake

neoblackcap:
Fix first feature file example mistake.


jenisys:

neoblackcap:
OK, I find several same spell "mistake" later. But I think the key words in document are same (all upper case or lower case) is better for behave beginner.

PR #297, opened 2015-03-11
step_registry: allow given <scenario> steps

aidan:
Sometimes it's a useful shorthand to refer back to previously
implemented scenarios as given steps for linear increments in
functionality

### DIFF
--- a/behave/model.py
+++ b/behave/model.py
@@ -680,6 +680,7 @@ class Scenario(TagAndStatusStatement, Replayable):
         self.was_dry_run = False
         self.stderr = None
         self.stdout = None
+        step_registry.registry.scenarios[name] = self
 
     def reset(self):
         '''
--- a/behave/runner.py
+++ b/behave/runner.py
@@ -164,6 +164,14 @@ class Context(object):
             # -- NOTE: Otherwise skipped if AssertionError/Exception is raised.
             self._mode = self.BEHAVE
 
+    def get_stack(self):
+        return self._stack[0]
+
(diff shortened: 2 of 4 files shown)


brianlucas:
Here's another way to write it via a step instead:

@step('I run the scenario "{scenario}"')
def step_impl(context, scenario):
  from behave import model
  for feature in context._runner.features:
    for current_scenario in feature.scenarios:
      if current_scenario.name == scenario:
        to_exec = ['{step_type} {name}'.format(
            step_type=s.step_type,
            name=s.name)
                   for s in current_scenario.steps]
        context.execute_steps(u'\n'.join(to_exec))
        break

aidan:
@brianlucas this step implementation preserves tables and multiline text:

```

@step('I run the scenario "{scenario}"')
def step_impl(context, scenario):
  for feature in context._runner.features:
    for current_scenario in feature.scenarios:
      if current_scenario.name == scenario:
          for step in current_scenario.steps:
              passed = step.run(context._runner, quiet=True, capture=False)
              if not passed:
                # -- ISSUE #96: Provide more substep info to diagnose problem.
                step_line = u"%s %s" % (step.keyword, step.name)
                message = "%s SUB-STEP: %s" % (step.status.upper(), step_line)
                if step.error_message:
                    message += "\nSubstep info: %s" % step.error_message
                assert False, message
          return

```

jenisys:
Sorry, will not be merged.
As @brianlucas stated there is a more generic way to do the same on your own.

PR #321, opened 2015-05-25
Rephrase Django integration, mention behave-django

bittner:
Make the existing integration options more prominent, and explain the differences among them.


jenisys:

PR #328, opened 2015-06-07
Allow failing features to be retried a specified number of times

MihaiBalint:
Quite frequently and annoyingly behave runs fail due to networking or other glitches that are independent of the test suite or the tested systems. This adds a cmd-line switch to configure a simple facility that repeatedly executes a failing feature. 
The failed attempts are not reported if a pass is obtained. If the feature still fails after the specified number of runs then behave will give-up and report the feature as failed.

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -292,6 +292,11 @@ options = [
                   file and the steps directory (instead of default path names).
                   """)),
 
+    (('-r', '--retry'),
+     dict(dest='retry_count',
+          help="""Retry failing features a number of times before declaring
+                  failure.""")),
+
     (('--stop',),
      dict(action='store_true',
           help='Stop running tests at the first failure.')),
@@ -506,6 +511,7 @@ class Configuration(object):
         summary=True,
         junit=False,
         stage=None,
+        retry_count=1,
         userdata={},
         # -- SPECIAL:
         default_format="pretty",    # -- Used when no formatters are configured.
--- a/behave/runner.py
+++ b/behave/runner.py
@@ -479,8 +479,11 @@ class ModelRunner(object):
                     self.feature = feature
                     for formatter in self.formatters:
                         formatter.uri(feature.filename)
-
-                    failed = feature.run(self)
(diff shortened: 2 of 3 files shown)


MihaiBalint:
Believe me, I understand what you are saying, I am already using retries in the step implementation :)

Here is my setup:
CI machine talks to server hosting tested webapp and to Saucelabs VMs that will open the webapp with various os/browser combos.

You can imagine that many things can go wrong in that setup (although most are timeouts with Saucelabs VMs).

Since I'm using BDD to assert functionality not reliability I thought it would be reasonable to retry failed features. I'll look into the behave runner suggestion to take this further.

Thanks for the review,
Mihai.

robertknight:
We have the same scenario as @MihaiBalint with intermittent connectivity failures when testing against Sauce Labs for behave tests that we run as part of regular automated system checks. The failure rate from the past 24 hours was about 5%. The comments in the first reply about removing inherent unreliability from a system are quite valid for unit and integration tests, but a small number of complete end-to-end tests are still a very useful thing to have and Behave is great for that.

Behave makes it easy to have retries within individual steps but as @MihaiBalint explained, with connections to remote browsers, it is necessary to retry at the scenario level. I'm currently proposing to do this by monkey-patching `Scenario.run` for tests with a specific tag: https://github.com/hypothesis/smokey/blob/sauce-reliability/smokey/features/environment.py#L49 .

jenisys:
Reopen to consider comment.

jenisys:
@robertknight
One suggestion, you should use ``scenario.effective_tags`` (instead of: ``scenario.tags``) in the ``before_feature()`` hook of "environment.py" in your example. This allows you to tag the feature with ``@autoretry`` and all scenarios will inherit that tag (and its behaviour) from the feature.

jenisys:
Added support based on @robertknight proof-of-concept example in module ``behave.contrib.scenario_autoretry`` that performs the monkey-patching when needed.

robertknight:
Thanks for the quick response :)

PR #331, opened 2015-06-26
Don't include skipped scenarios in junit reports if -k is specified

### LINKED ISSUE #330: Skipped scenarios are included in junit reports when -k is specified
Running behave with '--no-skipped --junit' produces a junit report with skipped scenarios

vrutkovs:
Fixes #330


jenisys:
Thanks, but note that I made additional changes/fixes:
* Reports for skipped feature are no longer created when --no-skipped is used
* Tests counting for scenarios with --no-skipped was not correct/logical 
  (skipped=0 but skipped scenarios were counted as tests)

PR #335, opened 2015-07-06
Fix execute_steps() example

miabbott:
When I tried to use `execute_steps()` as shown in the example, I was
getting an assertion about the steps needing to be in unicode.

`Assertion Failed: Steps must be unicode.`

It seems that the argument passed to `execute_steps()` must be a unicode
string otherwise it will not run.

This change updates the example for that method to show you must
provide a unicode string.

Signed-off-by: Micah Abbott <miabbott@redhat.com>


jenisys:

PR #338, opened 2015-07-20
docs/philosophy.rst: Correct to uppercase in example

charleswhchan:
Update case to make consistent with other comments.


charleswhchan:
@jenisys: Could you please review this PR?

Issue #308, opened 2015-04-05
Fix the documentation of execute_steps


PyDeq:
1) The documentation says: 

> Returns boolean False if the steps are not parseable, True otherwise.

This is not true. Looking at the code, there is no `return False` expression. It raises an exception in case of a failure (and then also there is also no need to `return True` otherwise).

2) The documentation says:

> If the execute_steps call fails (either through error or failure assertion) then the step invoking it will fail.

Actually, this is kind of true but very vague. The step implementation invoking it can automagically fail only by unhandling an exception (or by setting a "failed" status somewhere by `execute_steps`). I think that it would be more useful if you could tell the truth that `execute_steps` raises `AssertionError` in case of a failure so that the invoking step implementation can catch the exception and fail gracefully.

Or even better it can raise a `behave` specific exception that wraps the original exception raised by the sub-step (but then the runner would have to catch both `AssertionError`s and this exception; or all users would have to start raising this exception instead of `AssertionError`; or this exception can be a subclass of `AssertionError` (but I think that `execute_steps` raising an `AssertionError` makes sense only from the implementation point of view; from the design point of view, `execute_steps` does not necessarily make an assumption and then it does not make sense to expect `AssertionError`s; despite the fact that `except AssertionError` looks weird in the code)).

mixxorz:
Changing the behavior is a breaking API change. I know the docs weren't consistent with the behavior, but a people like you have figured out the actual behavior. If we change it now, we're going to be breaking tests.

For now, I'll open a PR to update the docs to reflect the actual behavior. In the future, we might want to look at PyDeq's suggestion and raise a behave specific exception.

jenisys:
stale-bot: disabled

PR #344, opened 2015-07-21
Docstring for execute_steps is now consistent with its behavior

mixxorz:
I updated the docs for `execute_steps` to be consistent with the method's behavior.

See #308


bittner:
@mixxorz Can you rebase this and we try to get it into `master`?

mixxorz:
Now in #609

Issue #320, opened 2015-05-12
Userdata is case-insensitive when read from config file


hashimmm:
For example, looking at the "debug on error" recipe, if "BEHAVE_DEBUG_ON_ERROR" is specified from the command line then it can be retrieved as `userdata.getbool("BEHAVE_DEBUG_ON_ERROR")` but if it is provided in the behave.userdata section of the config file it must be read as `userdata.getbool("behave_debug_on_error")`.

I am not sure what the original intent was but making both cases case insensitive will likely cause fewer problems to existing users, as opposed to making them both case sensitive.

jenisys:
I think you are mistaken. Both cases should be case-sensitive (AFAIK). 

Why do you think that the configuration-file section requires lower-case ?

NOTE:
Naming conventions for userdata are up to you (meaning each user or each project).
But I would use a consistent case-sensitive style. 

The example to which you are referring was using environment-variables initially before the user data feature was provided. Environment-variables often use UPPER-CASE style. Therefore, this example used it as well (and kept this convention even after it was rewritten to use the userdata-feature instead of the environment-var solution).

hashimmm:
What I meant to say is that the configuration file converts configuration variable names to lower case. 

This is because of the default behavior of python's [ConfigParser](https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.optionxform).

That is why, currently, when reading configuration provided in the configuration file it must be read as lower-case. This restricts the user to using lowercase names since userdata may be read either from the command line, in which case it is case sensitive, or the configuration file.

jenisys:
@hashimmm 
Thanks, I was not aware of that.
If that is a problem for you, you should use a JSON file for the user-data.

hashimmm:
@jenisys 
Right, thanks.

mixxorz:
We can make it case sensitive. I'll commit a PR to this. :smile: 

Sources:
http://stackoverflow.com/questions/19359556/configparser-reads-capital-keys-and-make-them-lower-case
https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.optionxform
http://stackoverflow.com/questions/1611799/preserve-case-in-configparser

jenisys:
INTEGRATED INTO: behave-1.2.6.dev0

PR #345, opened 2015-07-21
Fix config parser case sensitivity

mixxorz:
Hey there. I worked on #320. Config keys loaded from `behave.ini` should now be case-sensitive, similar to the behavior of user-data loaded from the command line.

I added a test for this functionality as well.

Let me know what you think. :grinning:


jenisys:
Thanks

PR #354, opened 2015-08-13
Fix #336 encode stacktrace in utf-8

### LINKED ISSUE #336: Stacktrace contents getting illegal characters inserted with text function.
Today we upgraded from 1.2.4 to 1.2.5 and we noticed that when a test fails our junit style reports contain invalid characters in the CDATA stacktraces.
We debugged it to this line:
model.py, line 1473 -> "error = _text(traceback.format_exc())"

The _text function seems to be new in this release and is mangling our stacktrace, inserting illegal characters.

If it is any extra value for reproduction, we are on Windows, with python 2.7.9.

## Here is a sample stack trace before it is run through the _text function:
Traceback (most recent call last):
  File "C:\\Users\\bXXXX\\Documents\\virtualenvs\\XXXX\\lib\\site-packages\\behave\\model.py", line 1456, in run
    match.run(runner.context)
  File "C:\\Users\\bXXXX\\Documents\\virtualenvs\\XXXX\\lib\\site-packages\\behave\\model.py", line 1903, in run
    self.func(context, *args, **kwargs)
  File "features\\steps\\installer_steps.py", line 210, in impl
    directories, task_names, reg_keys)
AssertionError

## Here is the resulting stacktrace, which when printed to the CDATA section of the report, produces an invalid report:

Traceback (most recent call last):
  File "C:�sersXXXX\Documentsirtualenvs\XXXX\lib\site-packagesehave\model.py", line 1456, in run
    match.run(runner.context)
  File "C:�sersXXXX\Documentsirtualenvs\XXXX\lib\site-packagesehave\model.py", line 1903, in run
    self.func(context, *args, **kwargs)
  File "features\steps\installer_steps.py", line 210, in impl
    directories, task_names, reg_keys)
AssertionError

We have worked-around the issue by patching calls to text and over-ridding the default encoding with "utf-8". So this is not a very important issue to see fixed, but we thought you would want to know. Hope this info helps! Bryan

arsenik:
#336 stacktrace contains file system paths which are not encoded correctly.

### DIFF
--- a/behave/model.py
+++ b/behave/model.py
@@ -1475,10 +1475,10 @@ class Step(BasicStatement, Replayable):
                 error = u'Assertion Failed: '+ message
             else:
                 # no assertion text; format the exception
-                error = _text(traceback.format_exc())
+                error = _text(traceback.format_exc(), 'utf-8')
         except Exception as e:
             self.status = 'failed'
-            error = _text(traceback.format_exc())
+            error = _text(traceback.format_exc(), 'utf-8')
             self.store_exception_context(e)
 
         self.duration = time.time() - start


arsenik:
I have a lot of ```OSError: [Error 193] %1 is not a valid Win32 application``` I wonder what is it.

jenisys:
@arsenik Thx, for your work
I integrated a slightly different solution in behave-1.2.6.dev0

PR #367, opened 2015-10-06
Add python 3.5 to the list of python versions for Travis

hockeybuggy:
This PR adds python 3.5 to the list of interpreters the tests are run against.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - 2.7
   - 3.3
   - 3.4
+  - 3.5
   - pypy
 
 cache:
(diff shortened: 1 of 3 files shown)


hockeybuggy:
It seems to be [failing here](https://github.com/hockeybuggy/behave/blob/6f5a8fe9da6b109cf83e4d3e960853331b03de81/test/test_runner.py#L180).

When I run it locally: 

```python
file = "/Users/douglas/devel/behave/test/test_runner"
info = "behave runner is masking context attribute 'thing' orignally set in stack_frame = traceback.extract_stack(limit=2)[0] (/Users/douglas/devel/behave/behave/runner.py:232)"
```

hockeybuggy:
It looks like its related to the new object `FrameSummary` in the [traceback module](https://docs.python.org/3.5/library/traceback.html#traceback.FrameSummary). That's as far as I have gotten so far.

hockeybuggy:
I don't think that this behaviour change is damaging. I removed the assertions. Feel free to weigh in if anyone thinks otherwise. :smile:

jenisys:
DONE (thanks for the work).
The assertion problems seem to be related with python 3.5.0.
The problem resolves itself in thin air if you use python 3.5.1 explicitly (or the newer nightly build).

jenisys:

hockeybuggy:
Thanks @jenisys

PR #370, opened 2015-10-19
--force-color commandline option for coloring output on terminals that formally do not support colors, such as Jenkins

mrkafk:


jenisys:
Why is the normal ``--color`` option not enough ?

mrkafk:
stream_supports_colors = isatty()
    self.monochrome = not config.color or not stream_supports_colors

`isatty` tests if output is real terminal. If not, it turns color off. However, some terminals (such as Jenkins job) may still cope with color while not being real terminals. This `isatty()` check prevents coloring behave's steps results in Jenkins. See this question and answer: http://stackoverflow.com/questions/30718995/behave-test-runner-has-no-colored-output-on-jenkins/33221229#33221229

smrpr:
This would be REALLY useful.

carlosmmelo:
I think theres already a plugin in jenkins for this: https://wiki.jenkins-ci.org/display/JENKINS/AnsiColor+Plugin

mrkafk:
@carlosmmelo Well that's a jenkins plugin for producing color output (in fact I've used it to test this pull request), but behave so far does not honor this output (precisely for reasons given above, that jenkins console is not real terminal and behave detects that).

toshipp:
any updates? this is very useful for me.

jenisys:
The solution should look slightly different. Instead of adding yet another command-line option, the type of the ``--color`` option should be changed that it allows to support an optional argument. This will allow you to force the color-mode to be on. In addition, the implementation should be backward-compatible to the existing behaviour.

EXAMPLE-SKETCH (proof-of-concept):
```python
# Explore how to use a three-state choice param with argparse:
#
#    program  --color        # USE: value=2 (auto, optional value).
#    program  --no-color     # USE: value=0
#    program  --color=off    # USE: value=0
#    program  --color=on     # USE: value=1  => FORCED_ON (stays ON)
#    program  --color=auto   # USE: value=2  => AUTO_DETECT (switch off is color-mode is not supported)

from __future__ import print_function
from enum import Enum
import argparse
import sys

class TermColor(Enum):
    off = 0
    on  = 1
    auto = 2

def parse_color(text):
    text = text.strip().lower()
    value = TermColor.__members__.get(text, None)
    if value is None:
        allowed = ", ".join(TermColor.__members__.keys())
        raise ValueError("INVALID COLOR: %s (allowed: %s)" % (text, allowed))
    return value

def main(args=None):
    """Provides an example how to use a three-state choice param
    together with argparse.
    """
    if args is None:
        args = sys.argv[1:]
    parser = argparse.ArgumentParser(description=main.__doc__)
    parser.add_argument("--color", dest="color", type=parse_color,
                        const="auto", nargs="?",    # Optional value
                        default=TermColor.auto, 
                        # choices=TermColor.__members__.keys(),
                        help="Specify color mode: auto, on, off")
    parser.add_argument("--no-color", dest="color", action="store_const",
                        const=TermColor.off, help="Use color=off")
    options = parser.parse_args(args)
    print("args: %r" % args)
    print("Option: color=%r;" % options.color)
    return 0

if __name__ == "__main__":
    sys.exit(main())
```

msabramo:
Merge conflict

aerique:
My life needs this feature.  When redirecting through `tee` I don't get colors even when using `--color`.

PR #375, opened 2015-11-28
Adds a "AroundScenario" hook

benwah:
This is an improvement suggestion; 

Currently there's no way to add a hook that will be guaranteed to execute regardless on whether or not a scenario succeeds or fails. If it fails, the after_scenario hook doesn't run.

I'm proposing to allow users to add a AroundScenario hook, in the form of a Context Manager.

### DIFF
--- a/behave/model.py
+++ b/behave/model.py
@@ -15,6 +15,18 @@ import time
 import traceback
 
 
+class DummyCtxManager(object):
+    def __init__(self, context, scenario):
+        self.context = context
+        self.scenario = scenario
+
+    def __enter__(self):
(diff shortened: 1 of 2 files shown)


jenisys:
Why do you think that the after_scenario hook is not called when a scenario fails ?
Why should we introduce a new concept instead of fixing the existing one (if that is necessary) ?

jenisys:
Fixture concept was added to behave that provides a similar concept in a more generic way.

benwah:
Sorry I didn't get back to you before.

Thanks for adding this.

PR #388, opened 2016-01-04
Importing range from six.moves fails - Fix

kiran6891:
Fix to make sure ImportError does not occur when trying to import range from six.moves.
If ImportError occurs, use the inbuilt python range method

### DIFF
--- a/behave/formatter/pretty.py
+++ b/behave/formatter/pretty.py
@@ -7,7 +7,13 @@ from behave.model_describe import escape_cell, escape_triple_quotes
 from behave.textutil import indent, text as _text
 import sys
 import six
-from six.moves import range
+
+# if importing range from six.moves fails
+# use python inbuilt range method
+try:
+    from six.moves import range
+except:
+    pass
 from six.moves import zip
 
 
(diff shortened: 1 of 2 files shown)


kiran6891:
AppVeyor build fail is not related to my patch.

jenisys:
Are you sure that is the right solution?
As I understand it, the problem is related to having a "too old version" of ``six``.
Therefore, updating ``six`` to a newer version should solve the problem.

kiran6891:
@jenisys Updating six to a newer did not solve the problem. Hence, had to make this change for my project. I guess many people would face the same problem going forward. Also, python has an inbuilt range method which works as needed for behave. Hence, proposed this solution.

jenisys:
Actually, I newer experienced this problem.

What platform and python version are you using ?
What six version are you using ?

kiran6891:
1. CentOS 6.5, Python 2.7.8
2. six == 1.10.0

jenisys:
Looks good.
But then I assume that you have a BAD PYTHONPATH, sys.path or something like that.
An idea would also be to setup a virtual environment as clean sandbox to verify and diagnose your problem.

Note I will not integrate this patch as long as I am not convinced that this is a behave issue.
Currently, it still looks to me like a user/user-environment problem.

jenisys:
CLOSED, should only be a problem in the user's workspace.

PR #391, opened 2016-01-08
run all examples in scenario outline by line number

cliffxuan:
currently when the line number of a scenario outline is specified, the closest scenario will be run whether that scenario being the first on in the example table of the same outline or the scenario above.

what i think would be useful is like the following:
1. if the line number of scenario outline is specified, all the examples in the scenario get run
2. if the the line number of an example in given, just run that example (this is the current behaviour already)
3. if a line between the line of scenario outline and the example table, run all the examples

then the downside my implementation is
1. it has introduced the dependency on `ScenarioOutline` in `runner_unit`.
2. result of `self.feature.walk_scenarios` is not cached.

feel free to suggest any better approach.

### DIFF
--- a/behave/runner_util.py
+++ b/behave/runner_util.py
@@ -14,7 +14,7 @@ import sys
 from six import string_types
 from behave import parser
 from behave.model_core import FileLocation
-
+from behave.model import ScenarioOutline
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


cliffxuan:
i realised this will not work as i expected, as i modified the code base on version 1.2.4 and some changes have been made into master since then, which expands the scenario outline into lines for the examples.

what i think would be useful is like the following:
1. if the line number of scenario outline is specified, all the examples in the scenario get run
2. if the the line number of an example in given, just run that example
3. if a line between the line of scenario outline and the example table, run all the examples

what's your opinion @jenisys ?

cliffxuan:
the updated code should do the trick :smile:

cliffxuan:
then the downside is
1. it has introduced the dependency on `ScenarioOutline` in `runner_unit`.
2. result of `self.feature.walk_scenarios` is not cached.

feel free to suggest any better approach.

jenisys:
Reopen

jenisys:
PROVIDED IN: behave-1.2.7.dev1
Similar, more generic functionality. 
It is now possible again to run all examples of ScenarioOutline by using a FileLocation in the ScenarioOutline (body/steps area).

PR #393, opened 2016-01-08
Add flag for randomization of the execution of tests

norapetrova:


jenisys:
This does not solve the complete problem.
There is a special mechanism in behave that ensures a Scenario with @setup tag is automatically run before another scenario if you use the file-location mechanism. 

```shell
# -- MAY RUN 1..3 scenarios: [@setup-scenario, selected-scenario, @teardown-scenario]
behave features/alice.feature:40
```

Similar, Scenario tagged with @teardown should be run last in the feature.
This is currently not ensured by using the implemented shuffle mechanism in the feature.

cliffxuan:
thanks for the reply. @setup and @teardown are really useful tags and should be taken care of! will come up with a fix for that.

the rationale for this feature is to expose dependencies between tests. our tests always run in the same order, and when we run them in different orders, some tests start failing. do you also consider it a useful feature?

KoenAchtergaele:
The conversation hereabove has ended rather abruptly January 8.
Are there still plans to merge this pull request with the main branch?
We too find this new functionality useful.

cliffxuan:
@KoenAchtergaele glad that you find it useful 😀

norapetrova:
@KoenAchtergaele I updated the code to take into account the comment @jenisys made. Unless there is anything else, I think it's ready to be merged.

louisnayegon:
@jenisys @bittner can you give me an expected time when this will be merged and released?

fredizzimo:
Randomization of the tests would be useful for me too

PR #423, opened 2016-03-21
sphinx.steps formatter: Support ref link for each step

ZivThaller:
adding a Cross-referencing arbitrary locations for each step.
enables the user adding a quick link to behave step!

```rst
# -- FILE: some_textfile.rst
use - :ref:`<step_name>`
```

ReST example:
```rst
For more details see step:  :ref:`When my step do "{something}"`.
```


jenisys:

ADAPTATIONS:
* step-label is now in front of heading (was: after step description)
* Ensures that ``docutils`` package is installed to enable the step-label functionality.

PR #429, opened 2016-03-31
Replace "Manual Integration" by "Automation Libraries" section

### LINKED ISSUE #323: Update Django Example to work with version >=1.7
Hello,
I was trying to create some tests for my django project, but it seems that the [script](https://github.com/behave/behave/blob/master/docs/django.rst) is deprecated for latest versions.

A fix was to remove the settings mechanism to the updated one,
Remove the existing lines,
```
from django.core.management import setup_environ
from myproject import settings
setup_environ(settings)
```
to,
```
import os
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "myproject.settings")
from django.conf import settings
```

then replace the
`from django.core.handlers.wsgi import WSGIHandler`
with the
`from django.core.wsgi import get_wsgi_application`

and finally the error I get is,
```
    raise AppRegistryNotReady("Apps aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```
Any ideas?
Thank you very much in advance!
Michael

ps:
When we finalise this, I will create a pull request to update this.

bittner:
- remove "Manual Integration" section to avoid updating code samples for newer versions of Django (heavy maintenance)
- add generic "Automation Libraries" section with simple code samples that should not require frequent changes or updates
- add new "Manual Integration" section with straightforward boilerplate code

Fixes #323.

### DIFF
--- a/docs/django.rst
+++ b/docs/django.rst
@@ -11,13 +11,15 @@ Django, though, varies slightly.
     Provides a dedicated management command.  Easy, automatic integration (thanks
     to monkey patching).  Behave tests are run with ``python manage.py behave``.
     Allows running tests against an existing database as a special feature.
-    `Setup instructions <https://pythonhosted.org/behave-django/installation.html>`_.
+    See `setup <https://pythonhosted.org/behave-django/installation.html>`_
+    and `usage <https://pythonhosted.org/behave-django/usage.html>`_ instructions.
 
 :pypi:`django-behave`
     Provides a Django-specific TestRunner for Behave, which is set with the
     `TEST_RUNNER`_ property in your settings.  Behave tests are run
     with the usual ``python manage.py test <app_name>`` by default.
-    `Setup instructions <https://github.com/django-behave/django-behave/blob/master/README.md#how-to-use>`_.
+    See `setup <https://github.com/django-behave/django-behave/blob/master/README.md#how-to-use>`_
+    instructions.
 
 .. _Django: https://www.djangoproject.com
 .. _LiveServerTestCase: https://docs.djangoproject.com/en/1.8/topics/testing/tools/#liveservertestcase
@@ -50,7 +52,13 @@ To start a web browser for interaction with the front-end using
 
 In your step implementations you can use the ``context.browser`` object to
 access Selenium features.  See the `Selenium docs`_ (``remote.webdriver``) for
-details.
+details.  Example using *behave-django*:
+
+.. code-block:: python
+
+   @when(u'I visit "{url}"')
+   def step_impl(context, url):
+       context.browser.get(context.get_url(url))
 
 .. _Selenium docs: http://selenium.googlecode.com/git/docs/api/py/api.html
 
@@ -72,7 +80,15 @@ To start a web browser for interaction with the front-end using
        context.browser = None
 
 In your step implementations you can use the ``context.browser`` object to
-access Splinter features.  See the `Splinter docs`_ for details.
+access Splinter features.  See the `Splinter docs`_ for details.  Example
+using *behave-django*:
+
+.. code-block:: python
+
+   @when(u'I visit "{url}"')
+   def step_impl(context, url):
+       context.browser.visit(context.get_url(url))
+
 
 .. _Splinter docs: http://splinter.readthedocs.org/en/latest/
 
--- a/docs/django.rst
+++ b/docs/django.rst
@@ -25,6 +25,45 @@ Django, though, varies slightly.
 .. _LiveServerTestCase: https://docs.djangoproject.com/en/1.8/topics/testing/tools/#liveservertestcase
 .. _TEST_RUNNER: https://docs.djangoproject.com/en/1.8/topics/testing/advanced/#using-different-testing-frameworks
 
+Manual Integration
+==================
+
+Alternatively, you can integrate Django using the following boilerplate code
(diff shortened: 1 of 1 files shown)


mixxorz:
This looks good! :+1:

bittner:
@jenisys [PyCon 7](https://www.pycon.it/en/) is in just a few days. Can we merge this and release an updated version of the docs? Would be nice to show that off on the Django track of the conference.

bittner:
@jenisys PLEASE merge this PR! There's really nothing that can go wrong with it.

bittner:
@jenisys I've read on [Google Groups](https://groups.google.com/forum/#!topic/behave-users/W0kgEB-1btg) that a 1.3.0 version is scheduled for release.

Can you please make sure this PR gets merged before that happens, so that we have Django-related documentation up-to-date? This is long pending.

Thank you! :+1:

jenisys:
Note that I added some comments/hints to the page.

bittner:
Just a quick feedback:

1. Section [Manual Integration](https://github.com/behave/behave/pull/429/commits/33fff2a) went missing. Would be nice if you could merge this in again. (I know the title of the PR is a bit misleading; I added that section just recently.)
1. I appreciate the [hints on the interaction patterns](https://github.com/behave/behave/blob/master/docs/django.rst#test-automation-libraries) you added. They are a bit hard to digest though (for both the content and the order -- it now looks like a part of the Splinter section, which it isn't).
  * Can we move this into a different or a separate chapter, and write just a sentence and link to there? (You maybe don't want to hear this, but Djangonauts will want to land here and find a hands-on setup guide, but little or even no theory, I believe.)
  * If we want people to preferably interact with the model layer we may need to give a detailed setup example (in addition to the `--stage` option).

Say, If I have a Django backend and an AngularJS frontend, what is the preferred way to test that my application works across the whole stack, i.e. from a user point of view. (This is what the client ordering the application implementation is interested in proving. Usually performed in a "test phase" with manual testing after implementation is completed..)

What if I have a simple Django website, and I want to test the template, views and models? How would the preferred test setup look like?

jenisys:
1. I will merge the "Manual Integration" part from above with the URL that you provided.
2. The hints were to prevent newcomers from focusing on the frontend / UI too much. There are so many "BDD examples" that do that (meaning they describe the HOW (you interact with UI) and not the WHAT). I added a similar hint (with less details) to the "more_info.rst" page that I just added recently.

To access the model layer directly, you need some kind of API (or interface), that can be used for such tests. If you have a REST API that is often it, because it should work often with the model layer. If you have another API (maybe only during development and not as part of the product), use it.

The core point is, you should not drive all your tests through the UI. It slows down and may make your tests brittle (because the UI changes often) and in addition causes many changes (at least in your test automation layer). Therefore, it is often better to split it up into subsystems were you replacing the missing parts with stubs/mocks using the Ports+Adapter (Hexagonal) architecture. This allows your test and test automation layer to remain more stable, but your implementation of the SUT/AUT may change.

The client is mostly interested (in my opinion):
* that the business rules (in the model layer) are implemented correctly (here is the money)
* that you are fast to respond to new features
* and I agree that the UI looks nice and is usable

But my focus is more on the technical side (and I know what works from my point of view). You are more client/customer-related from your job profile, at least in the django "problem domain". If you are to UI focused, it will slow you down when you have several iterations to add new features (and the UI changes every time).

bittner:
I appreciate the time and effort you put into explaining the details. Just a few notes:

* A hint like that is fine. Where it's **currently placed** though (at the end of the Splinter section) is not optimal. We should add a dedicated section maybe at the end of the page pointing the things out you explain.
* Let's come up with **better examples**. They should demonstrate how to do the WHAT (instead of the HOW), even with Selenium tests. -- I'm pretty sure this is a demanding task.

BTW, I totally agree with the [testing pyramid](http://martinfowler.com/bliki/TestPyramid.html) for the reasons you explain. Yet still, I can explain my top-down approach with TDD that BDD is a perfect fit for, but this may not be the right place here (see below instead).

## TL;DR

My [PyCon 7 talk](slides.com/bittner/pycon7-fix-price-projects-and-agile) should set the stage for a pragmatic discussion about testing. Cutting the story short, I'm helping projects that normally would do *no* testing at all leverage the powers of Gherkin for project management in a pragmatic yet still strictly Agile manner.

jenisys:
@bittner 
Better examples and a dedicated chapter sounds like a good idea. Make a proposal.
You may look at some of the notes to improve the cucumber documentation at "Cukenhagen", a Hackaton that was organised by the Cucumber team in Copenhagen approx. 1 month ago. 

SEE ALSO:
* https://github.com/cucumber/docs/issues 
  (note: Work notes and screenshots is in the issues, repository is not populated much)

P.S. 
I know the hint is not placed too good. We think where we move it later-on (or incorporate it into the new chapter).

bittner:
@jenisys Re chapter placement: As a quick fix you could place the hint as a final section of the document (after "Visual Testing", same hierarchy level).

Note that I explicitly state, _"With behave you can test anything on the Django stack: ..."_ (because for me behave is "just" a smart test runner). So, it's really about the code samples that should show off more than just Selenium tests (though that's where my approach starts and thus I have most of my experience  with).

In other words, a closing chapter summarizing the _"everything"_ from the introduction is something that should fit very well. Like a parable in literature.

PR #430, opened 2016-04-07
add tap formatter

jesg:
tap (test anything protocol) is a format to report test results.  it is
the test report format used by buildbot.

url: https://testanything.org/


mcepl:
Not that I would like to be too self-centered, but what about using some TAP-producing library, like (well, mine) https://pypi.python.org/pypi/bayeux ? It would allow you to add a way more YAML formatted data to the output.

Issue #428, opened 2016-03-30
Multidimensionality concepts domains


vit1251:
There are many possible Given, When and Then, and it would make sense to separate them on semantic group (by requirement document, version or another criteria) in the directory tree.

Let's I show you filesystem directory tree as exmaple:

```
 +
 +-- features
 |      +-- steps (with steps impl scope "global")
 |      +-- 01_user_authority.feature
 |      +-- 02_user_guest_main_page_abilities.feature
 |      +-- payment_operation
 |                  +-- steps (with steps impl scope "global" + "payment_operation")
 |                  +-- 03_user_payment_make_order.feature
```

this make implementation in separate scope "global" and "payment_operation" ... and have exception when try to use impl from level up scope ... it make possible to sure that no where then this scope no more use this implementation and does not trashing global scope of domain.

What you think about that?

jenisys:
First, I am not quite sure I understood why you think you need this directory structure. 

Second: In general, it is a good idea provide a hierarchical structure to counter to much complexity in one place. But this does not work with the current implementation (yet). You basically need an external or own runner implementation to use it for such a structure. I am working on a improved runner that provides a "scope concept", but it is not usable yet. The "scope concept" provides own scopes (or namespaces) for the ``steps`` directory or an ``environment.py`` file and you automatically inherit the parent scopes. Basically steps and type converters (and hooks) are registered in own namespaces.

vit1251:
> I am not quite sure I understood why you think you need this directory structure.

To separate different implementation scope i.e. "When I press button" may implemented as press button on "Android device" or "Android device emulator" it different implementation but similar scenarios.

P.S. Another benefits is run this scope by different parameters for example let's imagine in global scope:

```
  Given A few device 
      | Name    | Directory with implementation |
      | Android | android_impl                            |
      | iOS        | iOS_impl                                 |
      | WP9      | wp9_impl                                 |
   When I start scope "payment_operation"
   Then I have no error in this block
```

jenisys:
Mmh, that "Scenario" is executed only once, but you probably mean the right thing (Scenario Outline Examples table).

[...]

PR #433, opened 2016-04-14
Multiple step directories

aisbaa:
Hello,

As I've commented on [issue 428](https://github.com/behave/behave/issues/428), I need to be able to specify multiple step directories. Let me know if that is desired feature in behave community.

Usage:

```
$ behave --extra-steps ../global-steps/net --extra-steps ../global-steps/files
```

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -99,6 +99,12 @@ options = [
      dict(metavar="PATTERN", dest="include_re",
           help="Only run feature files matching regular expression PATTERN.")),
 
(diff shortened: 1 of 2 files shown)


jenisys:
Mmh, that is not the solution for this problem IMHO.

The solution that you provided can already be done by the user before the behave test-run by using the ``PYTHONPATH`` environment variable. A build-script can already do that with needing your change.

aisbaa:
Good point, thank you will try to solve my problem using PYTHONPATH.

Issue #416, opened 2016-03-10
JUnit report messages cut off


remcowesterhoud:
When generating JUnit reports at the end of a test it cuts off the message at 80 characters. It would be nice to see a config option for increasing the size of these messages.

Right now my report message looks like this:
Message: Timed out uploading file: C:\Jenkins\workspace\automated-test-ft2\resou...
It'd be nice to have the full message here, so I can instantly see what went wrong.

Pretty sure this the code that causes it. Located at /reporter/junit.py line 274
```python
message = _text(step.exception)
if len(message) > 80:
    message = message[:80] + "..."
```

jenisys:
As workaround for now, add another formatter to the "behave.ini" like (with formatter=plain or progress3, ...):

```ini
[behave]
format   = rerun
    plain
outfiles = rerun.txt
    report_plain.txt
```

If you need multiple behave test-runs instead of one, it will be better to provide this information on command-line to avoid overwriting the report files.

jenisys:
@bittner Can you look after that ?
I am not quite sure why this limitation was added.

bittner:
@jenisys The related code was introduced in [Feb 9, 2013 by you](https://github.com/behave/behave/commit/6ea27cc#diff-8f811b9e99d7b5e8c890e0fe51c2df23R211) yourself. The related tickets #63 and #69 don't mention details about this very detail. At least not for what I can see.

I would assume the 80 character limitation is an arbitrary measure to make the output less verbose. What if we always print out the whole stack trace?

jenisys:
That was what I was hinting.

jenisys:
FIXED IN: behave-1.2.6.dev0

PR #440, opened 2016-04-27
Don't truncate stack traces in jUnit reports

bittner:
Don't truncate stack traces in jUnit reports

Closes issue #416.


jenisys:

PR #441, opened 2016-04-27
Integrate hint on testing more nicely (add section "A Note on Testing")

### LINKED ISSUE #323: Update Django Example to work with version >=1.7

bittner:
Follow-up on #429. Closes #323

### DIFF
--- a/docs/django.rst
+++ b/docs/django.rst
@@ -25,6 +25,45 @@ Django, though, varies slightly.
 .. _LiveServerTestCase: https://docs.djangoproject.com/en/1.8/topics/testing/tools/#liveservertestcase
 .. _TEST_RUNNER: https://docs.djangoproject.com/en/1.8/topics/testing/advanced/#using-different-testing-frameworks
 
+Manual Integration
+==================
+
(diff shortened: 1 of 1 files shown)


jenisys:

PR #445, opened 2016-05-05
Force color formating

aisbaa:
Hi, I did some work on color formatting options. It's not finished yet, but I'd like to share my findings and get some feedback.

These changes partially solves my issue - it is possible to get color output even using pipes. Unfortunately PrettyFormatter assumes that it is writing to terminal if ``self.monochrome`` set to ``False``, which doesn't work well with pipes. Basically what I need:
- pretty formatter without terminal functionality
- or plain formatter with colors

## Proposal

- I'd like to create PlainColorFormater which would inherit from Plain (solves my problem).
- Later refactor PrettyFormatter to inherit from PlainColorFormater, because conceptually plain and pretty formatters do the same thing (unless there is a really really good technical reason why it should not be so).

formatter inheritance chain

Benefits of this approach:

- PrettyFormater only deals with terminal output.
- PlainColorFormater only adds color to PlainFormater.
- PlainFormater only deals with textual output.
- Potentially reduces code duplication.

## Notes

It is possible to use [unbuffer](http://expect.sourceforge.net/example/unbuffer.man.html) tool to fix this, though it is additional dependency that I really want to avoid.

Issue is reproducible with `behave --color=always features/background.feature | less -R`.

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -496,7 +496,7 @@ class Configuration(object):
     """Configuration object for behave and behave runners."""
     # pylint: disable=too-many-instance-attributes
     defaults = dict(
-        color=sys.platform != "win32",
+        color=TermColor.auto if sys.platform != "win32" else TermColor.never,
         show_snippets=True,
         show_skipped=True,
         dry_run=False,
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -8,7 +8,6 @@ import re
 import sys
 import shlex
 import six
-from enum import Enum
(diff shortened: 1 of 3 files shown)


aisbaa:
BTW, tests are failing because I've used Enum as suggested in https://github.com/behave/behave/pull/370 and it is not backported to older python versions. Suggestions on how to fix this issue are welcome.

aisbaa:
Apologies for this mess, will clean it up and submit another pull request.

PR #446, opened 2016-05-06
Remove Formatter scenario_outline(), examples() method (not used)

aisbaa:
It seems that `Formatter.scenario_outline` is not used, let me know if I'm wrong.


jenisys:

PR #448, opened 2016-05-10
Add status to scenarios in JSON report

remcowesterhoud:
When running behave with the json formatter no status was assigned to scenarios.
This change adds a method to the formatter which gets called at the end of a scenario (eos). The json formatter uses this to add the final status of a scenario to the json output.

### DIFF
--- a/behave/formatter/json.py
+++ b/behave/formatter/json.py
@@ -30,6 +30,7 @@ class JSONFormatter(Formatter):
         self.current_feature = None
         self.current_feature_data = None
         self._step_index = 0
+        self.previous_scenario = None
 
     def reset(self):
         self.current_feature = None
@@ -71,6 +72,10 @@ class JSONFormatter(Formatter):
             self.step(step_)
 
     def scenario(self, scenario):
+        if self.previous_scenario:
+            self.current_feature_element['status'] = self.previous_scenario.status
+        self.previous_scenario = scenario
+
         element = self.add_feature_element({
             'type': 'scenario',
             'keyword': scenario.keyword,
@@ -78,6 +83,7 @@ class JSONFormatter(Formatter):
             'tags': scenario.tags,
             'location': six.text_type(scenario.location),
             'steps': [],
+            'status': scenario.status
         })
         if scenario.description:
             element['description'] = scenario.description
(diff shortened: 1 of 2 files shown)


remcowesterhoud:
The tests are failing. Probably because the expected output had the status on a different line than the actual output.
I'd be happy to fix it but I am unable to run these tests locally. Whenever I do I'm getting this error: `ImportError: No module named parse`
I am 100% sure that I've got this module installed as I can use it without problem in other projects.

aisbaa:
Hi, it might be that you need to activate virtualenv before you can run feature tests.

```bash
> . .tox/py27/bin/activate
```

jenisys:
* Too many unnecessary whitespace changes.
* New method eos() is not needed in my opinion. This can be automatically detected elsewhere, like in other formatters. We do not change/break the formatter protocol for that.
* JSON formatter feature fails now according to CI (Travis)

remcowesterhoud:
@aisbaa Thanks, that was it. I'm not familiar with tox so I didn't know.

@jenisys Sorry, about those whitespace changes. I reformatted the code with my IDE which changed those.
I'll see if I can find another way to do this instead of adding the eos() method.

aisbaa:
@remcowesterhoud you could try using `formatter.scenario` method, it is called before each scenario, which means after previous scenario has been executed.

remcowesterhoud:
@aisbaa How am I gonna access the previous scenario object to request its status in this method though? Also this would skip setting the status of the last scenario from each feature so I'd also have to add a solution for this to the eof() method.

remcowesterhoud:
I could do something like this in the eof() method but surely there must be a better way

```py
# -- Update the status of all scenarios in this feature
        for element in self.current_feature_data['elements']:
            for scenario in self.current_feature.scenarios:
                if element['type'] == 'scenario' and element['name'].startswith(scenario.name):
                    if scenario.type == 'scenario_outline':
                        for outline in scenario.scenarios:
                            if element['name'] == outline.name:
                                element['status'] = outline.status
                    else:
                        element['status'] = scenario.status
```

aisbaa:
@remcowesterhoud you can try something like that 

```python
def scenario(self, step):
    if self.prev_scenario is not None:
        # save scenario status
    self.prev_scenario = step
```

aisbaa:
Good work :+1:

remcowesterhoud:
@aisbaa Thanks for the help :)

remcowesterhoud:
@jenisys Any update on this pr? Is it all good for merge this way or are there more flaws?

jenisys:
With minor adaptations.

PR #451, opened 2016-05-11
changed the scenario naming format to reference row attributes

christensonb:
changed the scenario naming format to reference row attributes

I want these two annotation schemas to work, and this change makes that happen.
annotation_schema = u"{name} -- @{row.id} {row[0]}"
annotation_schema = u"{name} -- @{row.id} {row[NAME]}"

### DIFF
--- a/behave/model.py
+++ b/behave/model.py
@@ -797,9 +797,8 @@ class ScenarioOutlineBuilder(object):
                 self.id = name      # pylint: disable=invalid-name
 
         example_data = Data(examples_name, example.index)
-        row_data = Data(row.id, row.index)
-        return self.annotation_schema.format(name=scenario_name,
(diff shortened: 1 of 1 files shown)


jenisys:
Why do you not use the column names as placeholders for row data ?

```gherkin
Scenario Outline: <Name> with age <Age>
   ...

  Examples
     | Name  | Age  |
     | Alice | 21   |
     | Bob   | 42   |

# LEADS TO:
#   Scenario Outline: Alice with age 21 -- @1.1 
#   Scenario Outline: Bob with age 42 -- @1.2
```

kyp76:
Hello
Stupid things but It seems you forgot the : after examples.

Let me know if I am wrong.

jenisys:
No response why this is needed => Close it.

PR #452, opened 2016-05-11
modified the function locator to allow for functions being decorated

christensonb:
This change will make it so I can decorate my step logic and still have Behave be able to report the original step logic.  Currently it will report the decorated step logic, which isn't helpful.

### DIFF
--- a/behave/model_core.py
+++ b/behave/model_core.py
@@ -170,8 +170,13 @@ class FileLocation(object):
         """
         func = unwrap_function(func)
         function_code = six.get_function_code(func)
-        filename = function_code.co_filename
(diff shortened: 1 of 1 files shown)


jenisys:
Could provide a motivating example why this is necessary ?

I tested your code and there was no difference to the existing implementation (due to the ``unwrap_function()`` call):

```python
...
@step( u'I use decorator 1')    # <== LINE_NUMBER: location.line
@given(u'I use decorator 2')
@when( u'I use decorator 3')
@then( u'I use decorator 4')
def step_decorated4(context):
         pass
```

jenisys:
No reasoning provided => Close it.

PR #469, opened 2016-06-29
Typo fix

ab9-er:
This should be definitely the silliest of the commits, but it was hurting my eyes to see that typo.


jenisys:

ab9-er:
👏

Issue #458, opened 2016-05-22
UnicodeEncodeError inside naked except block in __main__.py


mseery:
Hi, I'm running Behave 1.2.5 on OS X. Something is causing my Jenkins box to vomit, and it involves unicode text.

Long story short, if my test trips an exception which contains unicode in the message, Behave chokes on the exception and it is lost to the sands of time.

I have worked around this problem by removing utf8 text from the offending tests, but I would like to understand the root issue.

If the exception handling code inside runner.run() were more robust for this kind of text, I could figure out why this is gakking on my Jenkins box.

Thank you!

--mseery

```
Traceback (most recent call last):
  File "/usr/local/bin/behave", line 11, in <module>
    sys.exit(main())
  File "/Library/Python/2.7/site-packages/behave/__main__.py", line 123, in main
    print(u"Exception %s: %s" % (e.__class__.__name__, text))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019' in position 69: ordinal not in range(128)
```

The surrounding context:
```
    try:
        failed = runner.run()
    except ParserError as e:
        print(u"ParseError: %s" % e)
    except ConfigError as e:
        print(u"ConfigError: %s" % e)
    except FileNotFoundError as e:
        print(u"FileNotFoundError: %s" % e)
    except InvalidFileLocationError as e:
        print(u"InvalidFileLocationError: %s" % e)
    except InvalidFilenameError as e:
        print(u"InvalidFilenameError: %s" % e)
    except Exception as e:
        # -- DIAGNOSTICS:
        text = _text(e)
        print(u"Exception %s: %s" % (e.__class__.__name__, text))
        raise
```

jenisys:
@mseery
Sorry, you need to provide me a little bit more information to enable me to fix it:
* which exception was thrown and with which data (maybe: Comment out the text/print lines to get the ``reraise`` context)
* what is your environment setting (``LANG, LC_ALL, LC_CTYPE, LANGUAGE`` if defined)
* which Python version (I assume you have Python 2.x based on your stack traces)

jenisys:
Insufficient information.
Probably solved by pull #471 recipe.
PROBABLY FIXED IN: behave-1.2.6.dev0

PR #471, opened 2016-07-01
convert an object to unicode (py2) using __unicode__ method first

ftartaggia:
use case: text(AssertionError(u"some unicode string àèìòù")) raised a UnicodeError because it naively tried to call str() on a unicode object.

Since AssertionError defines its own __unicode__ method, it is convenient to delegate Unicode conversion to it.

Old text() behaviour is maintained for objects not defining a __unicode__ method

### DIFF
--- a/behave/textutil.py
+++ b/behave/textutil.py
@@ -75,8 +75,11 @@ def text(value, encoding=None, errors=None):
         # -- CONVERT OBJECT TO TEXT:
         try:
             if six.PY2:
-                data = str(value)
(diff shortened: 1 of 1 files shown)


ftartaggia:
see also issue #458

ftartaggia:
moreover, the hotfix branched from v1.2.5 tag, in case you're interested in shipping a hotfix version before v1.2.6 is ready

nattster:
+1 This fix works for me

jenisys:
@ftartaggia 
Integrated solution in: behave-1.2.6.dev0

PR #476, opened 2016-07-26
Fixed bool expresion on iter self.all_steps

ar45:


jenisys:
INTEGRATED FIX  INTO: behave-1.2.6.dev0
OOPS, thx for the catch. I fixed the iterator problem slightly differently.

Issue #478, opened 2016-08-03
Additional reporter


aisbaa:
Is it possible to register/use customized reporter without forking? If so - how?

jenisys:
No (and it is currently not planned).
If you want to use the ``Reporter`` API:

* provide a generic formatter class that provides the adaptation of the Formatter to the Reporter
* provide your own Reporter class (as workhorse for the output)
* provide your own Formatter class that delegates work to the instance of your reporter class
  by using the generic Formatter2Reporter

SKETCH:
```
from behave.reporter.base import Reporter
from behave.formatter.base import Formatter

class MyReporter(Reporter):
    ...

class Formatter2Reporter(Formatter):
      """Provides generic adapter implementation from Formatter to Reporter API."""

class MyFormatter(Formatter2Reporter):
      """Creates an instance of MyReporter and hands it to the base class to do the work.""" 
```

aisbaa:
Okay, as I understand bridging points would be:

- Formatter.eof -> Reporter.feature
- Formatter.close -> Reporter.end 

Is there a plan to merge formatter and reporter interfaces? It seems that both are responsible for runner state representation just different points in time.

jenisys:
Reporters are (currently) much more constraint and not so extensible in behave's internal usage.
One such concept should be sufficient. Currently, this is the ``Formatter`` because it contains support to use multiple formatters and you can define your own one.

aisbaa:
Thank you.

PR #480, opened 2016-08-05
Custom reporter

aisbaa:
Hello, I've implemented custom reporter feature that allows passing reporter class via command line using scoped class name.

```bash
$ behave -r behave.reporter.summary:SummaryReporter
```

I understand it is not in a perfect state yet, therefore feedback is greatly appreciated. There is one issue, I couldn't find how to produce similar [error message](https://github.com/behave/behave/blob/master/features/formatter.user_defined.feature#L117) to user-defined formatter:

```
error: format=<formatter.name> is unknown
```

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -127,6 +127,10 @@ options = [
                   formatter is used. Pass "--format help" to get a
                   list of available formatters.""")),
 
+    (("-r", "--report"),
+     dict(action="append",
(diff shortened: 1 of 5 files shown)


aisbaa:
Closing this pull request - since its possible to use formatter methods to achieve similar behavior to reporter.

https://github.com/behave/behave/issues/478

PR #522, opened 2017-01-20
Merge from behave-parallel

pi-sjp:


behave/formatter/plain.py
```diff
@@ -99,8 +100,9 @@ def result(self, result):
 
         status_text = result.status.name
         if self.show_timings:
```
xrg: I would prefer not to change the way results are displayed. However, we could exploit 'show_timings' to have another value that would also print the starting time. Sth. like:
if show_timings == 2:
 ...
WDYT?

behave/formatter/plain.py
```diff
@@ -99,8 +100,9 @@ def result(self, result):
...
-            status_text += " in %0.3fs" % step.duration
-
+            status_text = "started at %s, %s in %0.3fs," % (
```
xrg: I would prefer not to change the way results are displayed. However, we could exploit 'show_timings' to have another value that would also print the starting time. Sth. like:
if show_timings == 2:
...
WDYT?

pi-sjp: What, sorry, I didn't mean to merge this to behave main stream, must be a mistake.
This should stay in my private fork. 
I just find you have a "parallel support" pull request, that is great, how is that going?
I have been working on behave parallel for a while privately, adding some random stuff I need. Parallel run worked pretty good for me. 

Let me know if there is anything I can do.

Thanks
Sean

xrg: Well, printing the time each step has started at may be desirable. Your changes make sense.
But, OTOH, I would not try to push output formatting changes in the same context/PR as the parallel branch. The "plain" formatter might as well stay plain, uncluttered from extra information. Then, some "plain+" formatter could have some extra bells & whistles, you see.

PR #527, opened 2017-01-26
resolve #520 -- allow to set custom heading name in sphinx steps documentation

### LINKED ISSUE #520: Set a custom heading name for sphinx steps documentation
Currently, steps sphinx documentation heading is generated automatically from module name.
https://github.com/behave/behave/blob/master/behave/formatter/sphinx_steps.py#L224

But, final result is too long for me,
"package.module.module.module...steps.py"

Is there any way to set a custom name?

ivanprjcts:
This PR try to resolve issue #520 using the same solution is used by sphinx api-doc:

-E, --no-headings
 	Don’t create headings for the modules/packages

http://www.sphinx-doc.org/en/1.5.1/man/sphinx-apidoc.html#options

### DIFF
--- a/behave/formatter/sphinx_steps.py
+++ b/behave/formatter/sphinx_steps.py
@@ -108,8 +108,10 @@ class SphinxStepsDocumentGenerator(object):
 .. todo::
     Step definition description is missing.
 """
+    userdata_scope = "behave.formatter.sphinx"
     shows_step_module_info = True
     shows_step_module_overview = True
+    shows_heading = True
     make_step_index_entries = True
     make_step_labels = has_docutils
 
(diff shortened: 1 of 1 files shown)


jenisys:
@ivanprjcts 
Some comments after skimming over your code. We are not starting to add command-line option for every formatter variant. That does not scale, gets confusing and is not composable with user-defined formatters. 

As long as there is not better solution, use the userdata mechanism to configure your formatter. 
See how it can be done with userdata in [behave.reporter.junit.JUnitReporter](https://github.com/behave/behave/blob/master/behave/reporter/junit.py#L182).

ivanprjcts:
Could be this a better solution?

https://github.com/behave/behave/pull/527/commits/9058a674b79f6528ff54d2fb6d6f5864a99871be

PR #542, opened 2017-03-22
Unit test fix

kevbradwick:
Running the tests on a Windows file system results in an error for `TestConfiguration.test_read_file`. This change fixes it.


jenisys:
That differs between Python2 and Python3 versions.

jenisys:
Sorry, "Works for me" patches are not acceptable. Provide a fix that is acceptable for all python versions in the next case.

Issue #547, opened 2017-03-31
Crash when using step definition with optional cfparse parts


ftartaggia:
The scenario described below passes with behave 1.2.5, but breaks behave 1.2.6.dev0.

The problem is matchers.py's CFParseMatcher.__init__ method calling super(CFParseMatcher, self).__init__ instead of super(ParseMatcher, self).__init__

Scenario:
        Given a new working directory
        And a file named "features/steps/steps.py" with:
            """
            import parse
            from behave import step, register_type, use_step_matcher

            @parse.with_pattern(r'optional\s+')
            def parse_optional_word(text):
                return text.strip()

            register_type(opt_=parse_optional_word)

            use_step_matcher('cfparse')

            @step(u'some {:opt_?}word')
            def step_impl(context, opt_):
                pass
            """
        And a file named "features/alice.feature" with:
            """
            Feature: Alice
              Scenario: Bob
                Given some optional word
            """
        When I run "behave -f plain features/alice.feature"
        Then it should pass

jenisys:
INTEGRATED INTO: behave-1.2.6.dev0

PR #548, opened 2017-03-31
fix issue #547 "crash when adding a step definition with optional parts

ftartaggia:


pulphix:
When are we going to merge this fix?

jenisys:
NOTES:
* Provided fix in pull-request reintroduced implementation error from behave-1.2.5
* Correct, sane fix was provided that cleans up the problem of behave 1.2.5 and 1.2.6.dev0

PR #554, opened 2017-04-06
Adds galician language

carlosgoce:
Thank you


jenisys:
It would be better to add this to the language database, too (that is used to generate ``behave/i18n.py``:

* [bin/i18n.yml](https://github.com/behave/behave/blob/master/bin/i18n.yml)

There are some differences to cucumber galician language support:

* [cucumber/gherkin/gherkin-languages.json](https://github.com/cucumber/cucumber/blob/master/gherkin/gherkin-languages.json)

jenisys:
Add information to "bin/i18n.yml" based on [gherkin/gherkin-languages.json](https://github.com/cucumber/cucumber/blob/master/gherkin/gherkin-languages.json).

There was only one difference related to ``background``: ``Antecedentes`` was not contained. If you think that is necessary, add a corresponding issue or pull-request to https://github.com/cucumber/cucumber .

PR #567, opened 2017-06-02
add an option to fail on empty scenario outline

jeremad:
If someone write a scenario outline without any scenario, in a feature, the feature silently passes, some use may want to fail in this case.


jeremad:
I should probably add a test, though

PR #580, opened 2017-06-28
fixing some dead ecosystem links with some recent google results

smadness:
For broken links just did a google search and found what I thought would be the most relevant.


jenisys:

PR #584, opened 2017-07-09
Fix PendingDeprecationWarning

quitedensepoint:
Currently, a warning is emitted very frequently in console output when running Behave due to the outdated use of user_mode() instead of use_with_user_mode(). This fix addresses that problem.


jenisys:
Thx, already fixed.

PR #585, opened 2017-07-11
fix warning `Citation [TDD-Python] is not referenced.`

zhijiahu:
Got this error when building the docs with `tox -e docs`

_Warning, treated as error:
D:\behave\docs\more_info.rst:53:Citation [TDD-Python] is not referenced._

Fixed it by removing the reference in the Books section and followed how other references were added in Tutorials and Presentation Videos.

### DIFF
--- a/docs/more_info.rst
+++ b/docs/more_info.rst
@@ -50,12 +50,10 @@ Books
 
 `Behave`_ is covered in the following books:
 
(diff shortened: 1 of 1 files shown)


jenisys:
Should be already fixed.
I do not observe this behaviour w/ the current tip of the repository.

PR #588, opened 2017-08-01
Steps-catalog argument should not break configured rerun settings

Lego3:
We have a Behave configuration with the rerun formatter and a corresponding outfile. When we ran Behave with the --steps-catalog argument we realized that the rerun output file still got created, but with the steps-catalog output.

This pull request preserves the configured output file formatter coupling.

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -589,7 +589,10 @@ class Configuration(object):
         if self.steps_catalog:
             # -- SHOW STEP-CATALOG: As step summary.
             self.default_format = "steps.catalog"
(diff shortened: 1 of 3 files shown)


Lego3:
Any update on whether this fix can get merged into master? It is a quite small fix.

sirmackan:
I would also find this useful! +1

Lego3:
@jenisys, is there a possibility to get this merged?

Lego3:
This fix would still be useful.

Lego3:
The state is still the same. This fix would be useful.

PR #596, opened 2017-09-14
Add missing Czech translation

hason:
https://github.com/cucumber/cucumber/blob/master/gherkin/python/gherkin/gherkin-languages.json#L408


jenisys:
NOTE: Patched wrong file. The file is generated from the YAML file below (as file header states).
USE: [bin/i18n.yml](https://github.com/behave/behave/blob/master/bin/i18n.yml)

jenisys:

PR #599, opened 2017-09-27
Avoid Behave failure when step is in a different Windows drive

psicopep:
If for some reason you need to import steps from a different Windows drive Behave will fail because it tries to obtain a relative path. This simple fix just fallbacks to the absolute path in those cases.


jenisys:

PR #170, opened 2013-06-27
Support subfolders for step definition files

vrutkovs:
This change will allow to place step definitions in subfolders:

```
features/
+-- *.feature
+-- steps
|     +-- subfolder_1/
|     |    +-- *.py
|     +-- subfolder_2/
|     |    +-- *.py
```

etc.
Fixes #169

PR #600, opened 2017-10-06
config: options for path to environment file and steps directories

todfreed:
Makes it possible for behave to use steps and environment spread across
directories, rather than all being in the same tree

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -352,8 +352,14 @@ options = [
      dict(action="store_true",
           help="Show help for tag expressions.")),
 
-    (("--version",),
-     dict(action="store_true", help="Show version.")),
+    (('--version',),
+     dict(action='store_true', help="Show version.")),
+
+    (('--steps-directory',),
+     dict(action='append', dest='steps_directories', help="extra steps directories")),
+
+    (('--environment-path',),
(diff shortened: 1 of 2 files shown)


jenisys:
Sorry, that is the wrong way to go in my opinion.

It is already possible to use steps from multiple directories. You just need to import the other steps files in one file that is in  the steps directory. The other files must be reachable via the python search path.

SEE ALSO:
* [features/step.use_step_library.feature](https://github.com/behave/behave/blob/master/features/step.use_step_library.feature)

todfreed:
That doesn't cover this need, because you would have to know which files / steps to import. In contrast, to use steps from the steps directory, you don't have to do anything - behave finds them automatically.

If we don't have the autodiscover capability for multiple directory trees of steps in behave, then I have to implement the discovery of steps myself (and so do other users). This is the kind of thing that really should be implemented in behave proper.

jenisys:
Sorry, I disagree. You need to be in control when you import steps. In addition, you can solve this problem on your own. For example, extend the python search path in "features/environment.py" by using the userdata concept.

jenisys:
See comment on pull #170 (basic support provided for steps).

PR #614, opened 2017-12-20
Fix: More general type assertion

brolewis:
The type assertion was only allowing unicode when strings and unicodes
appear to be working fine.


jenisys:
Both values are of Unicode type. This is ensured by the Gherkin parser.
Therefore: Why is this a problem (for you) ?

brolewis:
So I updated my code locally with the code from this PR and moved on. I tried re-running directly from GitHub and it's running successfully so I'm closing. If I encounter the problem again, I'll try to see if I can include some test cases.

Issue #235, opened 2014-06-27
Run features (or scenarios) in parallel


julianpistorius:
I have many features and scenarios - mostly using behaving (https://github.com/ggozad/behaving) to run browser tests, so they're fairly IO-bound. 

Currently behave runs these in a serial fashion and it's taking a very long time (16 minutes and growing). I have many processor cores lying idle. How feasible is it to run these scenarios in parallel?

I've thought of a few hacky ways to do it, but ideally this would be a feature of behave itself.

Some ideas from other testing frameworks:
- https://github.com/grosser/parallel_tests
- http://www.opencredo.com/2013/07/02/running-cucumber-jvm-tests-in-parallel/
- http://nose.readthedocs.org/en/latest/plugins/multiprocess.html
- https://pypi.python.org/pypi/Scribbler/0.3.1

tomatobrown:
i would LOVE this!

iElectric:
+1

garlicPasta:
+1

doup:
+1

fedexcastro:
+1 any news about this?

tomatobrown:
looks like it's scheduled for the 1.3 release...

tomatobrown:
Is there any general idea of when the 1.3 release will be :)
:+1:

jronald01:
+1 for any general idea when this might be :)

phoebusliang:
+1

vishalm:
+1000

mubbashir:
+1

carlosmmelo:
+1

fedexcastro:
@jenisys, @julianpistorius, @tomatobrown guys, do you know if there is some advance on this feature?

tomatobrown:
I think jenisys will be the one to do this one, I'm actually considering a move to cucumber/ruby because it's already coming in that framework

julianpistorius:
@fedexcastro: No idea, sorry. We ended up rolling our own solution.

vishalm:
Hi Guys Any update on this issue !! and when is the plan for the release !! Let me know if I can help

vishalm:
Okey guys I figured out how to run behave tests parallel :) using behave-parallel. Let me know anyone want any help ... happy to help

tomatobrown:
Yes please, I'll start a ticket in that repo

SkyRG:
[...]

PR #616, opened 2018-01-03
parallel support

xrg:
This is a refactoring of tokunbo/behave-parallel, as in issue #235 . At this take, I have let the multiprocessing module fork *before* loading context or running any hooks. Then, the jobs are communicated using their `id()` (because of the fork, these mem. positions remain identical to the children) and then results are pushed back to the master (as primitive types, too). They are gathered in the master and "replayed" using the formatters/reporters.

This take seems to work well, and solves a problem where customized steps would not be pickle-able for tokunbo's approach.


xrg:
Tests for the parallel-feature fail on python3, because they use a `print` statement (rather than function). Can somebody help/advise on how to write them in an elegant, py2/3 compatible way?

spyoungtech:
@xrg the typical thing is to always use print as a function. To make this work for legacy Python2 like it does in *Python*, import the print function `from __future__ import print_function`. Just stick that at the top of the file then things should work as expected in both 2 and 3.

This is probably the easiest and most common way to address this problem. Hope that helps.

cmck:
hi @xrg 
I have been looking at your code on Windows 7 / 10, Python 3.6.3 and Python 2.7 against Chrome 63.0.3239.132 and get the following issue with pickle. Works fine on OS X. Could be related to this? https://bugs.python.org/issue18359

```$ behave --processes=3
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "c:\users\clark\appdata\local\programs\python\python36\Lib\multiprocessing\spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "c:\users\clark\appdata\local\programs\python\python36\Lib\multiprocessing\spawn.py", line 115, in _main
    self = reduction.pickle.load(from_parent)
TypeError: __new__() missing 1 required positional argument: 'line'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "c:\users\clark\appdata\local\programs\python\python36\Lib\multiprocessing\spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "c:\users\clark\appdata\local\programs\python\python36\Lib\multiprocessing\spawn.py", line 115, in _main
    self = reduction.pickle.load(from_parent)
TypeError: __new__() missing 1 required positional argument: 'line'
```

xrg:
@cmck Yes, you must be right, on a first glance. Let me play around with this and try to fix that issue..

xrg:
Bump. Any more views on that?

kageurufu:
The features in the multiprocessing runner do not have a parser object, so context.execute_steps fails

xrg:
@kageurufu Ack. Will try to deal with that

xrg:
Bump. Any comments?
I suppose the pypi build is failing because of 'multiprocessing' not working there, right?

xrg:
bump?

xrg:
Now works on pypy, too. :)

testingplusme:
@jenisys  Exactly. Also, I tested it and it looks like that it works. I think that is very necessary feature to have to parallel running of tests which can encourage more people to use behave in production projects.

xrg:
One issue: this patch /does/ break implied `pdb` support. The traceback is no longer stored as an object on the failed step, but rather the information of a traceback (formatted text) is. This is necessary to make that traceback serialisable across processes. Pdb will still work as a debugger, but need to instrument the exception deep there (we may need a different hook for that).

behave/model_core.py
```diff
@@ -299,9 +304,28 @@ def reset(self):
...
+        """
+        for key in 'exception', 'error_message', 'exc_traceback':
+            if key in value:
+                setattr(self, key, value[key])
+        if 'captured' in value:
+            self.captured.recv_status(value['captured'])
+
     def store_exception_context(self, exception):
         self.exception = exception
-        self.exc_traceback = sys.exc_info()[2]
```
sseliverstov: is it a necessary change? It broke allure reporter(

xrg: Necessary, as explained above. The tb object would not be serializable across processes. Might also benefit performance, because it frees any other objects bound to the traceback earlier on.

ntrepid8:
Is this PR being considered for a merge? Has it been reviewed or is anything blocking review? The diff isn't huge so what's holding it up?

Can those of use who want to see this merged offer any help to speed the process or reduce risk? Would it be helpful for people to test this on their own projects (as a few have already done) and post feedback?

naveenchandra91:
@jenisys I second @ntrepid8 on this. When can this PR be merged with base branch? Kindly let me know if I could be of any assistance with this.. We need to execute scenarios parallely to reduce execution time. This  changes would help a lot

Mdirfan25:
When is this getting merged ? or do we have any other solution in place for parallel execution ?

naveenchandra91:
@xrg: Please let us know if there is any reason due to which merge of this pull request been paused? Please let us know if you need any help with merging this with base branch.

xrg:
That is not in my end, we've all been waiting for project maintainers to 
consider this PR.


-- 
This email is free from antivirus checks.

xrg:
bump

jacekgaluszczynski:
We need this too. Is there anything we could do to support this getting merged?

johnprakashgithub:
Why it is still blocking for merge? Can someone solve the conflicts in it?

xrg:
There had been NO conflicts when the PR had been created *1 year* ago.

johnprakashgithub:
@xrg I have 5 features and I want each feature to take an item from the list when a parallel element is a feature. for example i have 3 elements in list=['a','b','c'] as global resource and 5 features. execute the features with " --processes 5 --parallel-element feature". first 3 features will take 3 elements f1 with a, f2 with b, f3 with c. Then f4 and f5 will be waiting till the acquired resource back in the list. f1 and f3 completes its execution and return a and c in the list ['a','c']. Now, f4 will pick 'a' and f5 will pick 'c'. After every feature completes its execution then the list=['a','b','c']. How can I do this in behave parallel?

glenn-barker:
@xrg Thanks for this, it's great and hope it gets merged in soon.

However, I've noticed a quirk where the final summary of my tests often result in "untested" rather than the expected "passed" or "failed" when I run them in parallel mode after installing your fork. Is this something you're aware of and is there any information I can provide to you to help investigate this further?

xrg:
@glenn-barker Not aware of such a thing, all runs in my case (using the default + junit formatters) have been reporting the stats properly. I'd be happy to investigate your setup and fix it, if needed.

xrg:
@johnprakashgithub Hehe, your case makes me think I could write a paper on it (already have some notes, indeed).
The short answer, however, is that you might be pushing re-usable resources back into the queue, as soon as they freed by some scenario. A multiprocessing queue would keep them in sync (hopefully).

glenn-barker:
> @glenn-barker Not aware of such a thing, all runs in my case (using the default + junit formatters) have been reporting the stats properly. I'd be happy to investigate your setup and fix it, if needed.

Sure, I have a simple example repo already here on GitHub that seems to reproduce this. Here's what I would run from start to finish to install my repo and run its tests with your behave-repo fork:

```
$ git clone https://github.com/glenn-barker/web-bdd.git
$ cd web-bdd
$ pipenv install
$ git clone -b eparallel https://github.com/xrg/behave-parallel.git
$ cd behave-parallel
$ pipenv run python setup.py install
$ cd ..
```

Then run with:

`$ pipenv run behave --processes 4 --parallel-element feature -D headless=true`

After a short time (my example repo only has a few feature files that do some basic navigation around Wikipedia) I get output like the following:

```
INFO: all sub-processes have returned
0 features passed, 0 failed, 0 skipped, 3 untested
0 scenarios passed, 0 failed, 0 skipped, 7 untested
0 steps passed, 0 failed, 0 skipped, 0 undefined, 24 untested
Took 0m0.000s
```

However if you examine the output (or turn `headless` mode to `false` in the param to kick things off) then you can definitely see that things are running as expected.

xrg:
@glenn-barker This pattern rings a bell: I've had cases before that results could not be serialized back to the master (see python pickle limitations) and hence counters would not record the numbers of tests passed. Tried hard to minimize the quirks (for one, that's the reason why steps cannot hold the traceback itself but only its text), but there may be leftovers that still cause this.
Do you notice anything other unusual in the logs (even by turning --no-capture to get direct stderr/stdout streaming) ?

xrg:
@glenn-barker Btw. https://github.com/xrg/behave_manners

glenn-barker:
@xrg Per your advice I've tried running using the following:

`$ pipenv run behave --no-capture --no-capture-stderr --no-logcapture --outfile BEHAVE_OUTPUT.txt --junit --processes 4 --parallel-element feature`

And observed the following:
* The resulting `BEHAVE_OUTPUT.txt` prints the feature tags, the feature names, the feature descriptions, and the feature background - but nothing related to scenarios or individual steps that execute.
* The `--junit` flag causes three xml files (one per feature) to be generated but they all indicate that the individual test cases were skipped / untested.

If I disable parallel execution by running with just:

`$ pipenv run behave --no-capture --no-capture-stderr --no-logcapture --outfile BEHAVE_OUTPUT.txt --junit`

I observe the following:
* The resulting `BEHAVE_OUTPUT.txt` prints the full output that you would expect, including the full scenario contents, every step, as well as a path to the step def file for each step.
* The three `--junit` xml files all indicate that the test suites passed with no failures and no skipped tests.

So it definitely seems to be "across the board", whenever this issue occurs then it seems that no reporting mechanism at all has any knowledge that these steps actually executed, as you alluded to.

Do you have any ideas what may be different in my setup that might cause this as opposed to when you don't see it? Even if this isn't something that can be solved I'd be curious if it could at least be mitigated by me configuring things a bit differently.

Let me know if any further info I can provide may help here.

xrg:
At Wednesday, 3 April 2019 5:23 Μ.Μ. glenn-barker wrote:
> @xrg Per your advice I've tried running using the following:

I may be able to explain once I setup an environment like yours and reproduce 
it here. At the moment I have not had time to set all this up, sorry.

-- 
This email is free from antivirus checks.

johnprakashgithub:
> @johnprakashgithub Hehe, your case makes me think I could write a paper on it (already have some notes, indeed).
> The short answer, however, is that you might be pushing re-usable resources back into the queue, as soon as they freed by some scenario. A multiprocessing queue would keep them in sync (hopefully).

@xrg I tried multiprocessing queue and found that always every process takes its own queue copy. Multiprocessing queue I initiated in before_all as context.queue = Queue() and then used "get" from queue in before_feature and "put" to queue in after_feature. However, this doesn't work, all the features are getting the same element in the queue.

johnprakashgithub:
it's not working in the following code when I run it's in parallel
implementation tried as 2 feature files f1.feature and f2.feature. Then the environment.py contains the following code 
from multiprocessing import Queue

queue = Queue()
queue.put("first")
queue.put("second")

def before_feature(context, feature):
    feature.new_element = queue.get()
    print("before_feature of {} the element from queue is {}".format(feature.name, feature.new_element))

def after_feature(context, feature):
    print("after_feature of {} the element to queue is {}".format(feature.name, feature.new_element))
    queue.put(feature.new_element)

run the feature files as 

behave --process 2 --parallel-element feature features/f1.feature features/f2.feature 

i get the result as 
before_feature of first feature the element from queue is first
before_feature of second feature the element from queue is first
after_feature of first feature the element to queue is first
after_feature of second the element to queue is first

xrg:
@johnprakashgithub You're right, I was able to reproduce your case yesterday. 
Longer explanation to follow up later tonight..

xrg:
@johnprakashgithub 
So, the longer explanation: `environment.py` is loaded by the `load_hooks()` method of the runner, late in the forking process. Reason is simply to keep the method implementation of the original runner, which loads its hooks just before invoking the loop.
Likewise, the root context is only created on the (child) runner. I did that deliberately, because putting any non-pickleable object in it would block forking of the sub-process.
These, combined, mean that both global variables of `environment.py` and `before_all()` hook are invoked locally on the children workers, not early on the master one. Which is a problem for sharing resources, agreed. On the other hand, this minimizes interference between workers.

I am thinking of a good way to provide a resource-sharing API, without moving `before_all()` or the root context up to the master process (which would make the patch more intrusive against current non-parallel code).

A workaround now (which I don't particularly like, however) is to have some other python file imported by both the `environment.py` and any of the step implementation. Reason is, step implementations are loaded early in the feature-scanning process (on the master) and then any referenced module will end up in `sys.modules` and participate in forking. Then, `environment.py` referencing it will import an already existing module, its globals could contain the shared queue in that case.
But is not pretty, I admit. Wonder where the 'fixtures' concept would help here, doing that import in a more elegant way..

johnprakashgithub:
> @johnprakashgithub
> So, the longer explanation: `environment.py` is loaded by the `load_hooks()` method of the runner, late in the forking process. Reason is simply to keep the method implementation of the original runner, which loads its hooks just before invoking the loop.
> Likewise, the root context is only created on the (child) runner. I did that deliberately, because putting any non-pickleable object in it would block forking of the sub-process.
> These, combined, mean that both global variables of `environment.py` and `before_all()` hook are invoked locally on the children workers, not early on the master one. Which is a problem for sharing resources, agreed. On the other hand, this minimizes interference between workers.
> 
> I am thinking of a good way to provide a resource-sharing API, without moving `before_all()` or the root context up to the master process (which would make the patch more intrusive against current non-parallel code).
> 
> A workaround now (which I don't particularly like, however) is to have some other python file imported by both the `environment.py` and any of the step implementation. Reason is, step implementations are loaded early in the feature-scanning process (on the master) and then any referenced module will end up in `sys.modules` and participate in forking. Then, `environment.py` referencing it will import an already existing module, its globals could contain the shared queue in that case.
> But is not pretty, I admit. Wonder where the 'fixtures' concept would help here, doing that import in a more elegant way..

Thanks, @xrg for such a detailed explanation. I will do a deep dive on fixtures. I too thinking of a way to provide an API for resource sharing during parallel execution.

johnprakashgithub:
@xrg Also think of what are the problems in merging it to masters. I do see that model, main and init have conflicts. However, I don't have write access.

johnprakashgithub:
@xrg I was trying to implement sharing resources by using the fixture.  The same example as earlier I was trying and getting the same problem that before_feature always picks all the workers first item in the queue. 
Above each feature I provided @fixture.queue.myqueue
then in environment.py

from behave import fixture, use_fixture
from multiprocessing import Queue

@fixture
def my_queue(context):
    context.queue = Queue()
    context.queue.put('rg-816')
    context.queue.put('rg-803')
    yield context.queue
    # context.queue.task_done()


def before_tag(context, tag):
    if tag == "fixture.queue.myqueue":
        use_fixture(my_queue, context)

johnprakashgithub:
def before_feature(context, feature):
    feature.new_element = context.queue.get()
    cp = multiprocessing.current_process()
    print("Multiprocessing id {} before_feature of {} the element from queue is {}".format(
        cp.name,
        feature.name, feature.new_element))


def after_feature(context, feature):
    cp = multiprocessing.current_process()
    print("Multiprocessing id {} after_feature of {} the element to queue is {}".format(
        cp.name,
        feature.name, feature.new_element))
    context.queue.put(feature.new_element)

xrg:
In the meantime, I remembered from my notes why I'd chosen /not/ to share the `before_all()` method, but rather to call it on each subprocess separately:
`before_all()` needs the context. I could have initialized a context and tried to keep it during the fork(). But then, there should be an `after_all()` which might do any corresponding cleanup to the context. At that stage, there would be multiple copies (forks) of that original context, rather impropable to merge back to a single instance. Or calling `after_all()` multiple times against that context, breaking the symmetry..

Then, if I would add a `before_everything()` method, to be called once at the master process, that would leave the non-parallel mode working a different way. 

I'm looking for a solution for `behave_manners`, as well. It has to be some way to define fixtures that would be instantiated on the master, copied through the fork and shared thereafter. I may need to add some code to `runner_mp.py` but also on the standard `runner` to achieve that.

johnprakashgithub:
@xrg Can you provide a sample implementation?

xrg:
> @xrg Can you provide a sample implementation?

There you are my friend: in 'tmp-resources' branch of my 'behave' repo I have setup an experiment and managed to share your queue to the subprocesses. But needed to add a line to 'runner_mp.py' to achieve that.

johnprakashgithub:
> > @xrg Can you provide a sample implementation?
> 
> There you are my friend: in 'tmp-resources' branch of my 'behave' repo I have setup an experiment and managed to share your queue to the subprocesses. But needed to add a line to 'runner_mp.py' to achieve that.

@xrg Thanks 👍 , Now the shared resources are working fine across features while using tmp-resources.

lulapm:
I executed some tests in parallel using this solution and ran successfully. The problem is that test results output aways informe the features untested. How resolve this problem? Do you working in this improvement?

johnprakashgithub:
@lulapm I do see that problem in some specific scenario when I use sample codes. However, I do not see any problem in nightly execution with parallel implementation. Right now I have new test implementation that is where I spend most of the time. I will come back to it later.

LogicalDash:
@lulapm I think I fixed formatting for your use-case, please try again?

levi-rs:
I would also like to see this functionality added to Behave if possible.

xrg:
Bump. I keep waiting on this PR.

behave/runner_mp.py
```diff
@@ -0,0 +1,266 @@
...
+            job = self.jobs_map.get(job_id, None)
+            if job is None:
+                print("ERROR: missing job id=%s from map" % job_id)
+                self.jobsq.task_done()
+                continue
+
+            if isinstance(job, Feature):
+                yield job
+                try:
+                    self.resultsq.put((job_id, job.send_status()))
```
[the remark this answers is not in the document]
xrg: It's a tuple put in the queue

behave/runner_mp.py
```diff
@@ -0,0 +1,266 @@
...
+                for k in ('filename', 'line', 'keyword', 'name', 'tags',
+                          'description', 'background', 'language'):
+                    kwargs[k] = getattr(job.feature, k)
+                kwargs['scenarios'] = [job]
+                orig_parser = job.feature.parser
+                feature = Feature(**kwargs)
+                feature.parser = orig_parser
+                yield feature
+                try:
+                    self.resultsq.put((job_id, job.send_status()))
```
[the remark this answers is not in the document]
xrg: it's a tuple, again

Vadmeggy:
Bump

cariel27:
Bump

karpaivan:
Bump

xrg:
As it stands, the only solution is to abandon the official branch for this one. After all, this one does NOT break compatibility with 1.2.6

akuma12:
Also bump. Would love to see this implemented. I tried it out on my existing features and it worked great.

xrg:
s/implemented/merged
The implementation seems to be stable now, after almost 2 years of scrutiny.

Keep an eye on my repo, please: since my projects being tested grow, along with behave and `behave-manners`, I'm now really considering adding extra functionality to `behave` itself. That is, to go beyond Gherking language spec and support more sophisticated testing strategies.

behave/__main__.py
```diff
@@ -180,7 +181,33 @@ def main(args=None):
...
+                print ("ERROR: When using --processes, --parallel-element"
+                    " option must be set to 'feature' or 'scenario'. You gave "
+                    "'%s', which isn't valid." % pelem)
+                return 1
+
+        except ImportError as e:
+            print ("DEBUG: import error: %s" % e)
+            print ("ERROR: Cannot import multiprocessing module."
+            " If you're on python2.5, go get the backport")
+            return 1
```
bittner: OMG, how old is that code!?! :astonished: Do we really need to support Python 2.5?

behave/__main__.py
```diff
@@ -180,7 +181,33 @@ def main(args=None):
...
+                rclass = MultiProcRunner_Feature
+            else:
+                print ("ERROR: When using --processes, --parallel-element"
+                    " option must be set to 'feature' or 'scenario'. You gave "
+                    "'%s', which isn't valid." % pelem)
+                return 1
+
+        except ImportError as e:
+            print ("DEBUG: import error: %s" % e)
+            print ("ERROR: Cannot import multiprocessing module."
```
bittner: There shouldn't be spaces between `print` and the opening parenthesis. Please run `flake8` over the files to find issues like that.

behave/__main__.py
xrg: @bittner b500c39535b4c17605ec , from 2013!
Still, it would make sense if run on some py interpreter that doesn't support multiprocessing today.

kingbuzzman: @xrg out of curiosity, what interpreter would that be? And if so, couldn't they just install https://pypi.org/project/multiprocessing/ (very outdated) instead of having a fork in the code to address a very unlikely scenario, plus extra nesting?

xrg: TBH, i don't really know/remember why that 'if' would be needed for multiprocessing. I had just refactored that old code years ago. And /thought/ that there were still some mis-behaving interpreters around. Removing the conditional could classify as an improvement, agreed.

kingbuzzman: Agreed then. Ps. can you merge upstream/master down and fix your conflicts?

johnprakashgithub: I

> OMG, how old is that code!?! 😲 Do we really need to support Python 2.5?

Python 2.5 support we shouldn't be worried now. Fix it by removing the if condition and at least get it running in 2.7 would make sense.

xrg:
Note: the original request was against 1.2.6, and still applies perfectly against that. Unfortunately, the head of 'behave' has moved, adding new features and breaking compatibility, without advancing the version number. For that head, I'm working on another branch, which merges and considers new features (see `parallel-130` in my repo).

varac:
Please get this merged soon !

xrg:
Yay! We got over the 2 years mark with this!! 🎂

karpaivan:
How we (I) can help to see PR in 1.2.7?

derrix060:
Any ETA for people review it?  Having parallel support helps a lot!

seibmj:
Is there anything that is needed to get this to ready besides resolving the conflicts.  Team is having to go off the older [behave-parallel ](https://github.com/tokunbo/behave-parallel/releases).  Trying to decide to just abandon the master and point to this branch instead of using the old behave parallel .

xrg:
@seibmj There is not much to do (on my side) about this branch: it applies *perfectly* to the behave version it has originally been developed against (1.2.6). But perhaps we need another 2-3 years before it gets reviewed!
On the other hand, I do have a branch where 'parallel' is merged with the breaking 1.2.7 changes, and that one works too:  https://github.com/xrg/behave-parallel/tree/parallel-130 .. Too early for that one, I'd suppose..

bittner:
>  But perhaps we need another 2-3 years before it gets reviewed!

Thanks for the polemics!

How about resolving the conflicts? I hope you understand that it's not possible to merge this PR without this being addressed.

dani29:
As a new user of **behave** (and a guest in this repo), the saga around concurrency in behave is pretty confusing and clearly some people feel frustrated. I am hoping not to come across as too judgemental, but just wondering if there's any timeframe for this being resolved - merge conflicts, reviews, etc.? Seems that this feature is very interesting for quite some users.

will-yangsg:
`objc[83919]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[83919]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
ob`

Does anybody run into this problem on mac?

xrg:
3 years!! and counting...

pl-aknight:
FWIW: pytest-bdd + pytest-xdist

bittner:
Good to know some people haven't yet got tired of counting. :+1:  I'd suggest you rebase your feature branch and resolve the conflicts? We'll then take a look and make sure this gets merged (or the problem domain gets addressed in some way or the other).

Thank you for your understanding! ... and Happy New Year! Good things will happen. :four_leaf_clover:

irtikapoor:
Any timeframe for this being resolved ? I am also struggling with this, and wants to run the behave test in parallel ?

yogevn-c2a:
Is there any progress with this PR?

darcyc-chen:
Hi @xrg,
I have a  problem when I run scenarions in parallel and use  allure formatter report at the same time.
allure formatter result was not generated when I use command:
`D:\python_project\behave-parallel-master\bin\behave -f allure_behave.formatter:AllureFormatter -o re
sult --processes 4`
And the result was generrated well if I use command:
`D:\python_project\behave-parallel-master\bin\behave -f allure_behave.formatter:AllureFormatter -o re
sult `
 
any solution about this?

jordan-gillard:
> Hi @xrg, I have a problem when I run scenarions in parallel and use allure formatter report at the same time. allure formatter result was not generated when I use command: `D:\python_project\behave-parallel-master\bin\behave -f allure_behave.formatter:AllureFormatter -o result --processes 4` And the result was generrated well if I use command: `D:\python_project\behave-parallel-master\bin\behave -f allure_behave.formatter:AllureFormatter -o result `
> 
> Any solutions about this?

Does the _Support behave parallel_ section of [the allure-behave package docs](https://pypi.org/project/allure-behave/) answer your question?

It suggests updating `environment.py` files (or creating one if you don't have one already) and explicitly adding the Allure reports directory. i.e.

```python
# environment.py
from allure_behave.hooks import allure_report

### your code

allure_report("path/to/result/dir")
```

wiebren:
@withnale I / Claude had a stab at this; https://github.com/wiebren/behave/commit/26b07b380ffe0d1c1f0612e12261ec579666ea20

treat as highly experimental, when I have some more confidence I may make a PR request out of it to upstream it

wiebren:
> @withnale I / Claude had a stab at this; [wiebren@26b07b3](https://github.com/wiebren/behave/commit/26b07b380ffe0d1c1f0612e12261ec579666ea20)
> 
> treat as highly experimental, when I have some more confidence I may make a PR request out of it to upstream it

now available as a separate package; https://github.com/wiebren/behave-parallel-runner

PR #620, opened 2018-01-31
combinations: specify scenario outline examples as the product of sets

todfreed:
This patch enables a new way of specifying the examples of a scenario
outline, in addition to listing them exhaustively in an Examples block.

With a Combinations block, the scenarios are built up from the
combinations of sets of parameters. This makes it possible to express a
large number of scenarios.

Each line in a Combinations block specifies a parameter group, which
includes one or more parameter names, and one or more lists of values
for those parameters. If a parameter group contains more than one
parameter, then the names are separated by commas, and the value lists
are enclosed in parentheses.

Testing:
Scenario Outlines with one or more Examples, and one or more
Combinations groups work as expected. Regular Scenarios continue to work
as before. pytest passes.


todfreed:
https://github.com/behave/behave/pull/621

PR #621, opened 2018-01-31
combinations: specify scenario outline examples as the product of sets

todfreed:
This patch enables a new way of specifying the examples of a scenario
outline, in addition to listing them exhaustively in an Examples block.

With a Combinations block, the scenarios are built up from the
combinations of sets of parameters. This makes it possible to express a
large number of scenarios.

Each line in a Combinations block specifies a parameter group, which
includes one or more parameter names, and one or more lists of values
for those parameters. If a parameter group contains more than one
parameter, then the names are separated by commas, and the value lists
are enclosed in parentheses.

Testing:
Scenario Outlines with one or more Examples, and one or more
Combinations groups work as expected. Regular Scenarios continue to work
as before. pytest passes.


jenisys:
Interesting idea, but I need to think about it if I accept it in the current form.

PROBLEMS:
* Gherkin needs to be extended (and we should remain compatible w/ Gherkin syntax IMHO).
* BAD: Your table syntax for "Combinations" varies from normal table syntax.

The functionality can be provided w/ a slightly different approach:
Provide an **Example tag/fixture** that serves as **Example Table generator** (or table transformator).
The "fixture" function provides the table generation / transformation algorithm and is executed before the scenarios are generated from the scenario outline.

ADVANTAGES:
* Behave model only needs to be changed slightly 
  (basic functionality needed to support Example Table transformation / Example Table generation from Base table)
* Can be used for other Example Table generations approach, too (pairwise, permutations, reading example table from a file, ...)

```Gherkin
Scenario Outline: Use 
   Given I use <Stuff>
   And I use <Thing>
   ...

  @table_generator.combinations
  Example: Generate Example Table by using combinations of parameters
    | Stuff | foo, bar |
    | Thing | baz, qux |
```

NOTE: Actually, you want to use [pairwise](http://pairwise.org) instead of "combinations". Otherwise, your parameter-space will explode with additional parameters (parameter dimensions) and needs too much time for executing all combinations.

hossbeast:
No, combinations is exactly what I want. I understand the potential for combinatorial explosion of test cases, but for exhaustive testing, this is what you want.

The pairwise concept is interesting though. I can see how it would be useful to specify the sets of parameters and values in the same way, but to do "pairwise" generation of test cases, instead of exhaustive generation. Permutations would be yet another way of generating test cases from the same sets of parameters and values.

What about specifying those mechanisms like this?

```
Permutations:
  Stuff  | foo, bar |
  Thing | baz, qux |
```

```
Pairwise:
  Stuff  | foo, bar |
  Thing | baz, qux |
````

The table parsing would be just the same as for a Combinations table, the generation algorithm would be a bit different.

jenisys:
One of the core ideas should be that table-generators or table-transforms can be provided by user (extending the one the behave provides). Therefore, independent what behave provides, it can be solved w/o change of behave (after the core mechanism is provided).

PR #627, opened 2018-02-28
Fix UnicodeDecodeErrors in Python 2

rgonalo:
Avoid UnicodeDecodeError in Python 2 in two cases:

* when captured logs contain utf-8 characters
* when a Windows host configured in a non-ascii language raises an exception with non-ascii chars

### DIFF
--- a/behave/_types.py
+++ b/behave/_types.py
@@ -1,6 +1,7 @@
 # -*- coding: UTF-8 -*-
 """Basic types (helper classes)."""
 
+import platform
 import sys
 import six
 if six.PY2:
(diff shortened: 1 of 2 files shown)


behave/_types.py
```diff
@@ -55,8 +55,15 @@ def has_traceback(cls, exception):
...
         # -- NORMAL CASE:
-        text = u"{prefix}{0}: {1}\n".format(exception.__class__.__name__,
-                                            exception, prefix=prefix)
+        try:
+            text = u"{prefix}{0}: {1}\n".format(exception.__class__.__name__,
+                                                exception, prefix=prefix)
+        except UnicodeDecodeError:
+            # Avoid UnicodeDecodeError in Python 2 when 'exception' is a system
+            # error thrown by a Windows host configured in a non-ascii language
+            exc_message = str(exception).decode('cp1252', 'ignore')
```
jenisys: That is a Windows specific solution that you applied to all platforms. This may bite other users on other platforms or with other Windows encoding / charset settings.

rgonalo: This is of course a Windows specific solution, but it does not affect to other users, as UnicodeDecodeError is not handled in behave base code and they would only get an exception in any case.
Anyway, I updated it not to mask the original exception for non-Windows users.

jenisys:
@stalebot
Remains open.

PR #642, opened 2018-04-06
Fix: issue 631: Outline variables not possible in table headings

mschnelle:
Replace Outline Variable also in Table Headings.


jenisys:
INTEGRATED INTO: behave-1.2.7.dev0
Thx

PR #649, opened 2018-04-27
Fix #648 Incorrect tutorial code sample

### LINKED ISSUE #648: Debug-on-Error tutorial code sample incorrect
The _Debug-on-Error (in Case of Step Failures)_ sample code from behave/tutorial.rst is incorrect since `step.status` returns an Enum and not a string, hence the user will never be able to perform debug upon a step failure since the comparison is invalid.

https://github.com/behave/behave/blob/master/docs/tutorial.rst#debug-on-error-in-case-of-step-failures

hateno:
Fix for #648


jenisys:
You are wrong. The example is still correct. The ``Status`` enum class has a comparison operator that accepts strings. Therefore: ``assert True is (Status.failed == "failed")`` does not fail !!!

PR #656, opened 2018-05-17
wip: Add support for 3rd party reporters

BernardoSilva:
- [ ] allow developers to setup their custom reporters from config file

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -429,6 +429,7 @@ def read_configuration(path):
     # SCHEMA: config_section: data_name
     special_config_section_map = {
         "behave.formatters": "more_formatters",
+        'behave.reporters': 'more_reporters',
         "behave.userdata":   "userdata",
     }
     for section_name, data_name in special_config_section_map.items():
@@ -578,6 +579,7 @@ class Configuration(object):
         self.environment_file = "environment.py"
         self.userdata_defines = None
         self.more_formatters = None
+        self.more_reporters = None
         if load_config:
             load_configuration(self.defaults, verbose=verbose)
         parser = setup_parser()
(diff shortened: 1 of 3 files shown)


behave/configuration.py
```diff
@@ -673,6 +677,17 @@ def setup_formats(self):
             for name, scoped_class_name in self.more_formatters.items():
                 _format_registry.register_as(name, scoped_class_name)
 
+    def setup_reports(self):
+        """Register more, user-defined reporters by name."""
+        print('current reporters:', self.reporters)
```
veryhappythings: Is print a good choice here? Can we use a logger instead?

behave/reporter/_registry.py
```diff
@@ -0,0 +1,120 @@
...
+        raise ValueError(message)
+    module_name, class_name = parse_scoped_name(scoped_class_name)
+    reporter_module = load_module(module_name)
+    reporter_class = getattr(reporter_module, class_name, None)
+    if reporter_class is None:
+        raise ImportError("CLASS NOT FOUND: %s" % scoped_class_name)
+    return reporter_class
+
+
+def select_reporter_class(formatter_name):
```
veryhappythings: `formatter_name` should be `reporter_name` I think? Same throughout the method

behave/configuration.py
```diff
@@ -673,6 +677,17 @@ def setup_formats(self):
...
+        if self.more_reporters:
+            for name, scoped_class_name in self.more_reporters.items():
+                reporter = _report_registry.load_reporter_class(scoped_class_name)
+                if not _report_registry.is_formatter_valid(reporter):
```
veryhappythings: This should be `is_reporter_valid`, I think?

jenisys:
One extension-point for output-reports is enough and that is currently for formatters.
Therefore, I will not put any effort into making reporters also an extension-point.

HINT:
You can use your own reporter class and use an adapter-class to register it as formatter.

PR #657, opened 2018-05-21
Allow async steps with timeouts to fail when they raise exceptions

ALSchwalm:
Currently, when an async step is decorated with `async_run_until_complete(timeout=...)` all exceptions  (aside from the timeout) are suppressed. So a step like:

    @then("we do an async thing")
    @async_run_until_complete(timeout=1)
    async def step_impl(context):
        assert(False)

Will not fail. Note that the step will fail if the 'timeout' parameter is not given. This PR checks the completed async task for an exception and propagates it if there is one.


jenisys:
INTEGRATED INTO: behave-1.2.7.dev0

Issue #662, opened 2018-06-06
setup.py fails on Python 3.3


jayvdb:
tox.ini says behave supports 3.3, but out of the box `setup.py` fails

```
$ python setup.py -q install
/opt/python/3.3.6/lib/python3.3/distutils/dist.py:258: UserWarning: Unknown distribution option: 'python_requires'
  warnings.warn(msg)
error in behave setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Expected version spec in traceback2; python_version < '3.0' at ; python_version < '3.0'
```

https://travis-ci.org/jayvdb/behave/jobs/388593320

jayvdb:
c.f. https://github.com/behave/behave/issues/661 which also needs to be fixed if py33 is to be supported.

jenisys:
Sorry, I think you are mistaken. This does not look like a behave problem. This looks like a setuptools problem. Use a current version of setuptools / pip and you will not get this problem. 
NOTE: ``python_requires`` support was added in "v24.2.0" (2016-07-20). I assume that you have an older version installed.

SEE ALSO:
* https://setuptools.readthedocs.io/en/latest/history.html

PR #663, opened 2018-06-06
Re-enable Python 3.4 on Travis

### LINKED ISSUE #661: Python 3.4 on Travis
I tried to do a build on Python 3.4, as I am investigating a problem on that version in a @coala project. c.f. https://github.com/coala/coala-ls/issues/5

I had a similar problem locally here.  It seems there are a few problems with `six` 1.10, which is probably why behave uses `six>=0.11`.  But on Python 3.4, that isnt enough to bypass the problems.  easy_install seems to be still installing 0.10

```
Searching for six>=1.11.0
Reading https://pypi.python.org/simple/six/
Downloading https://files.pythonhosted.org/packages/16/d8/bc6316cf98419719bd59c91742194c111b6f2e85abac88e496adefaf7afe/six-1.11.0.tar.gz#sha256=70e8a77beed4562e7f14fe23a786b54f6296e34344c23bc42f07b15018ff98e9
Best match: six 1.11.0
Processing six-1.11.0.tar.gz
Writing /tmp/easy_install-ml1c7_vh/six-1.11.0/setup.cfg
Running six-1.11.0/setup.py -q bdist_egg --dist-dir /tmp/easy_install-ml1c7_vh/six-1.11.0/egg-dist-tmp-52rvhofa
no previously-included directories found matching 'documentation/_build'
zip_safe flag not set; analyzing archive contents...
__pycache__.six.cpython-34: module references __path__
creating /home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/six-1.10.0-py3.4.egg
Extracting six-1.10.0-py3.4.egg to /home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages
Adding six 1.10.0 to easy-install.pth file
Installed /home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/six-1.10.0-py3.4.egg
```

https://travis-ci.org/jayvdb/behave/jobs/388576411

jayvdb:
Fixes #661

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,13 @@ python:
   - "3.6"
   - "2.7"
   - "3.5"
+  - "3.4"
   - "pypy"
   - "pypy3"
   - "3.7-dev"
 
 # -- DISABLED:
 #  - "nightly"
-#  - "3.4"
 #  - "3.3"
 #
 # NOW SUPPORTED: "3.5" => python 3.5.2 (>= 3.5.1)
--- a/py.requirements/ci.travis.txt
+++ b/py.requirements/ci.travis.txt
@@ -1,3 +1,4 @@
+six >= 1.11
 mock
 nose
 PyHamcrest >= 1.9


jayvdb:
tbh, this isnt a true fix of #661, as it only works around the problem.

Likely some combination of pre-installing a more recent pip & setuptools would also work.
I'm happy to do that instead, if preferred.

That would likely still only workaround the problem - a 'real fix' would be for setup.py to require that a recent pip/setuptools is installed, but that depends on a recent setuptools (catch22) and causes other breakages, such as seen at https://github.com/behave/behave/issues/662

jenisys:
Sorry, you are fixing things with a side-effect instead of fixing the problem. ``six >= 0.11`` is automatically installed when you install behave.

jayvdb:
@jenisys , are you suggesting that you would prefer for behave to specify a minimum version of `setuptools` that is needed on py34 ? (assuming setuptools is where the problem is)

As mentioned earlier, I am happy to provide a patch for that approach instead, if preferred.

jenisys:
Just check your setuptools version and check the features in the setuptools history that behave is using in "setup.py".

Setuptools had an approach how to ensure that setuptools is installed and has a valid/newer version.
But this approach is deprecated now. Therefore, behave cannot specify a minimal version except as comment.

pip (and setuptools) are your packaging infrastructure in python.
Ensure that they are sufficiently new and you will not run into problems.
For example, you should use ``pip >= 10.0`` since the new ``pypi.org`` was enabled (March/April 2018).

PR #666, opened 2018-06-16
add # noqua: F811 to schema of undefined step

squallcs12:
When adding undefined step in to steps  files, it raise flake8 error as the method name are all the same. So I would like to add a flake8 ignore for undefined step schema

### DIFF
--- a/behave/runner_util.py
+++ b/behave/runner_util.py
@@ -431,7 +431,7 @@ def make_undefined_step_snippet(step, language=None):
     if single_quote in step.name:
         step.name = step.name.replace(single_quote, r"\'")
 
-    schema = u"@%s(%s'%s')\ndef step_impl(context):\n"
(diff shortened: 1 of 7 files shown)


jenisys:
Are you sure that is really needed?

NOTE: This is a convenience feature to make your life easier (but not perfect for everybody).
And you need to tweak the code anyway.
And can provide it with unique, sensible names ;-)

jenisys:
I normally use my keyboard to make the step name unique.
This allows me to directly call step implementations, too.
NOTE: And I implement my steps (which cannot be done by anybody else).

Step snippets are a convenience tool.
Provide a more "sophisticated paste-method" in your favourite IDE or editor (as macro, etc.) and the problem is solved for you.
NOTE: Snippets will never be right for everybody (like you and me), because other people use other tools with other conventions.

Issue #630, opened 2018-03-10
class-based step implementations


spyoungtech:
Currently, step implementations are provided as module-level functions. I am proposing that behave should support the ability to provide step implementations as methods in a class.

I strongly feel that class-based definitions would be a great enhancement for behave. 
Some of the possible benefits of this feature includes:

- Provides a defined pattern understood by both step library writers and users
- A more flexible way for step library writers to provide steps to users, allowing easy extension of previous works
  - users can subclass library steps to extend/override its methods, providing new ways for behave users to solve unique challenges
  - users can pick & choose what step definitions get into the registry
- A simple and beautiful, yet powerful interface to behave internals
  - Keep signatures cleaner, so `context` need not be the first argument of every step
- Easier to write tests for your tests! (e.g. easier dependency injection via inheritence)
- Less direct reliance on global state, while maintaining compatibility
- (possibly) Make it easier for IDE plugins to detect definitions from step libraries
- Solves some [obscure problems](https://github.com/spyoungtech/behave-problem-example) for step library writers
- And much more

I hoping to discuss this idea here.

# Proposal draft; one possible interface

The following is a conceptual sketch of just one possible interface for this vision. There are a lot of ideas baked in here that, for brevity sake, are not totally explained. Naturally, there are a lot of details that would need to be worked out to implement something like this. For now, use your imagination and assume that obstacles to achieving this can be overcome :)

## A base class or mixin

Behave may provide some base class or mixin containing some behave-specific logic and internal interfaces.

```python
# behave/step_registry.py ?
class StepImplMixin(object):
    step_matcher = 'default' # use default defined in environment.py or parse
    _local_registry = LocalStepRegistry()
    


    @property
    def context(self):
        """proxy for retrieving the current context"""
        # some smart logic
        return ctx

    def register(self, method_names=None):
        """
        Inserts methods specified in ``method_names`` from the local registry to the global step registry.
        if ``method_names`` is not specified, all methods from the local registry will be inserted
        (maybe register definition patterns with (an instance of?) this class?)
        """
        ...


    # probably other cool stuff, maybe unittest assertion matchers?
```

Then a library writer (or even users) could provide step definitions something like

```python
# some_library/steps/some_steps.py
class SomeSteps(StepImplMixin, OtherClass):
    step_matcher = 'parse' # override the default matcher
    # this decorator would put the method into the class-local registry
    @given_impl(u'I do "{something}"') # write a given step using the class-level matcher
    def given_something(self, something):
[...]

PR #681, opened 2018-08-02
class-based step definitions

spyoungtech:
This PR aims introduces class-based step definitions as a feature to `behave`. The primary goal of this feature is to introduce a beautiful way of extending upon existing code in step definitions; beyond what it already available with `context.execute_steps`. It is primarily aimed at developers of reusable step libraries, but has clear benefits for everyone.

This feature would provide the following:

- ability to define step definitions as classes
- ability to extend steps from your own classes (or perhaps classes provided by other libraries/packages)
- ability to define a matcher per-method without changing global state of the 'current matcher'
- wraps methods to transform `context` into an attribute (`self.context`), so it's not necessary to have each method use `context` as first parameter in the signature. Works with default runner.

These basics are covered in the added test (`features/class_steps.feature`). 

Additionally, I believe class-based definitions would make it easy to introduce a lot of other niceties. For example, to get assertion matchers from the stdlib `unittest` module (e.g. `self.assertEquals`) one only has to use `unittest.TestCase` as a mixin to the class: 
(extending on example in `features/class_steps.feature`)
```
import unittest
from base_steps import SomeSteps

class ExtendedSteps(SomeSteps, unittest.TestCase):
    def step_then_I_have(self, amount, item):
        actual = self.context.shopping_cart.get(item, 0)
        self.assertEquals(actual, amount, 
            f'Expected {amount} of "{item}", but cart had {actual} of "{item}"'
        )

ExtendedSteps().register()  # register local definitions to the global step registry
```

This and many other integrations are made possible and simple by class-based definitions. I hope to provide some more compelling examples, but I believe at a minimum, this change will help users make step implementations that are more flexible and DRY compliant.

Notes:

- There's definitely refactoring potential in `LocalRegistry` class, if some elements are moved to the `StepRegistry` class -- I chose to leave `StepRegistry` alone for the sake of making the PR more palatable to merge in as an experimental, purely opt-in, feature

- per-step matchers are more or less an independent feature: they could easily be adapted to existing StepRegistry/decorators, with or without the rest of the changes.

- current implementation only considers the default behave runner; could probably be adapted to be more flexible with other runners.

Related issue: #630 

Would love to hear thoughts on this. If this has a remote interest in being added to `behave`, I'll be happy to write end-user documentation for this feature.

### DIFF
--- a/behave/runner.py
+++ b/behave/runner.py
@@ -20,8 +20,8 @@ from behave.formatter._registry import make_formatters
 from behave.runner_util import \
     collect_feature_locations, parse_features, \
     exec_file, load_step_modules, PathManager
-#from behave.step_registry import registry as the_step_registry
 import behave.step_registry
+
 if six.PY2:
     # -- USE PYTHON3 BACKPORT: With unicode traceback support.
     import traceback2 as traceback
@@ -610,8 +610,6 @@ class ModelRunner(object):
         # pylint: disable=too-many-branches
         if not self.context:
             self.context = Context(self)
-        # if self.step_registry is None:
-        #     self.step_registry = the_step_registry
         if features is None:
             features = self.features
 
--- a/behave/runner.py
+++ b/behave/runner.py
@@ -20,7 +20,7 @@ from behave.formatter._registry import make_formatters
 from behave.runner_util import \
     collect_feature_locations, parse_features, \
     exec_file, load_step_modules, PathManager
-import behave.step_registry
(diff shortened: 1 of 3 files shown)


spyoungtech:
I decided it may be more appropriate to provide this functionality as a 3rd party package. So, I've used these changes to create `behave-classy`, which has been released to PyPI.

Feel free to reopen or ping me if there's still an interest in having this merged.

PR #693, opened 2018-09-25
Setup log file - add filename handling in setup_logging()

marcinlesek:
Hello,

last weeks I faced some issues with setting up logging in our tests and I noticed, that `behave` not allow us to set logging to file feature. We need to work around it by manually specify `logging.basicCofnig()` by our own and use `logging`, but after checking how `behave` is handling it on his site, I saw that is using this same.

This little feature allows us to use and setup log file for our tests execution logs, but it's not necessary, so it had **backward compatibility**.

Also, where I could find some contribution guide? I couldn't get any information about the PR template, necessary checklists etc.

Best,
Marcin

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -557,6 +557,7 @@ class Configuration(object):
         self.junit = None
         self.logging_format = None
         self.logging_datefmt = None
+        self.logging_filename = None
         self.name = None
         self.scope = None
         self.steps_catalog = None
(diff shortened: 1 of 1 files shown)


jenisys:
@marcinlesek 
Did you solve the problem or did look at the [features/logging.setup_with_configfile.feature](https://github.com/behave/behave/blob/master/features/logging.setup_with_configfile.feature) ?

* Logging to a file works
* Use the documented examples
* Logging setup is tricky (and I tried to make it better to use)
* Some logging setup functions can only be executed once, like `logging.basicConfig()`.
  You can call them many times but only the first call does what you expect.

marcinlesek:
@jenisys Yea, I dig more in `behave` code and found, that it could work. Also I found that my contribution is useless regarding `**kwargs` usage, because it handles it. 
Regarding documented examples, I didn't find any regarding logging into the file.

Best,
Marcin

SamuelePilleri:
I ran into the same problem and found this thread. Apparently there's no easy way to log to a file, how am I supposed to do?

My current `behave.ini` contents are the following:
```ini
[behave]
logging_format=[%(levelname)-8s] %(asctime)-10s - %(message)s
logging_level=DEBUG
log_capture=False
```

Thank you

PR #718, opened 2018-12-27
update mock's link to the standard library page

deronnax:
mock has been integrated in python standard library with Python 3.3,
released in 2013. References to mock should point to that one rather
than the kept-for-compatibility PyPI package


deronnax:
Hello. Is there something wrong with my change ?

deronnax:
Could someone please merge it so it goes off my dashboard, please ? It's not even a code change and it's a link to official documentation. If not I'll closed it within few hours

Issue #719, opened 2019-01-03
Broken links under behave/doc


ki4070ma:
I found some broken links in [behave/docs/gherkin.rst](https://github.com/behave/behave/blob/master/docs/gherkin.rst).

E.g.
* .. _`Python step implementations`: tutorial.html#python-step-implementations
* .. _`environmental controls`: tutorial.html#environmental-controls
* .. _`configuration file`: behave.html#configuration-files

These should be rst, not html.
I didn't check other rst files.

jenisys:
See reason in #720.

PR #720, opened 2019-01-03
Fix broken links in doc

ki4070ma:
https://github.com/behave/behave/issues/719


jenisys:
REJECTED, the hyperlinks must work in the HTML documentation and not in Github's cross-referencing mode of `*.rst` files.

Issue #746, opened 2019-04-25
Store error and captured logging separately in the model


lawnmowerlatte:
When debugging failed tests, if the captured logging is verbose, you can easily lose sight of the error since it is printed above the captured logs. I would like it if the error were reprinted at the bottom of the captured logs. I would implement this as a formatter, but as far as I'm aware this can only be done as part of the model code. I have implemented the change that would store the two error types separately so a formatter would be able to use them as it saw fit.

lawnmowerlatte:
Any update here?

PR #747, opened 2019-04-25
Store error and logging separately in the model code so formatters can access it

lawnmowerlatte:
When debugging failed tests, if the captured logging is verbose, you can easily lose sight of the error since it is printed above the captured logs. I would like it if the error were reprinted at the bottom of the captured logs. I would implement this as a formatter, but as far as I'm aware this can only be done as part of the model code.

I have implemented the change that would store the two error types separately so a formatter would be able to use them as it saw fit.

### DIFF
--- a/behave/model.py
+++ b/behave/model.py
@@ -1660,7 +1660,7 @@ class Step(BasicStatement, Replayable):
                 self.captured = runner.capture_controller.captured
                 error2 = self.captured.make_report()
                 if error2:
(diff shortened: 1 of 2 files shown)


lawnmowerlatte:
Fix for #746

lawnmowerlatte:
Any update?

lawnmowerlatte:
Why?

PR #751, opened 2019-05-30
gherkin: Adding Rule keyword translation in portuguese and spanish to gherkin-languages.json

dunossauro:
Fix correct translation keyword in Portuguese and Spanish

### DIFF
--- a/behave/i18n.py
+++ b/behave/i18n.py
@@ -5,7 +5,7 @@
 # pylint: disable=line-too-long, too-many-lines, missing-docstring, invalid-name
 """
 Gherkin keywords in the different I18N languages, like:
- 
+
 * English
 * French
 * German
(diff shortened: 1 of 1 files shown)


jenisys:
@dunossauro 
Yes, but sadly this is the wrong place (and you are not the first with this "problem" ;-)). 
The python module is generated from a configuration file.

Check with the original file in the cucumber repository shown below.
Provide a pull-request there, then I will update and regenerate the python module here.

SEE ALSO:
* https://github.com/behave/behave/tree/master/etc/gherkin
* https://github.com/cucumber/cucumber/blob/master/gherkin/gherkin-languages.json

dunossauro:
@jenisys 
Thanks for the report that. In the same day I opened a PR to cucumber too.

I'll put keywords on
* https://github.com/behave/behave/blob/master/etc/gherkin/gherkin-languages.json

The last question, should I keep the changes in `i18n.py`?

jenisys:
INTEGRATED INTO: behave-1.2.7.dev1 (using  merged cucumber pull-request #621)
Thx

Issue #767, opened 2019-08-01
use_fixture_by_tag doesn't return fixture if fixture registry has bare functions


jgentil:
Right now if you use `use_fixture_by_tag` and your registry is bare functions (instead of function/args/kwargs tuples) the fixture itself isn't returned from `use_fixture_by_tag`.   But it is returned if the triple tuple is present.  

Lines in question:
https://github.com/behave/behave/blob/c2ef9aa02334e97455d826cdb38e8e9a96b84ea2/behave/fixture.py#L273-L279

jenisys:
MERGED INTO: behave-1.2.7.dev1

PR #768, opened 2019-08-01
FIX: use_fixture_by_tag didn't return the actual fixture in all cases

jgentil:
Fixes issue https://github.com/behave/behave/issues/767


jenisys:
INTEGRATED INTO: behave-1.2.7.dev1
Thx

PR #780, opened 2019-10-03
The string of Python 3 doesn't need `u` prefix

sih4sing5hong5:
For reader-friendly, removing the prefix `u` in Python 3.


jenisys:
As long as Python2 is supported this will remain as is:
* Python3 has no problem with it
* You are free to remove the 'u' for your steps (if you are using only Python3)

PR #787, opened 2019-10-22
Fix junit JUnit XML output. Fixes #510

### LINKED ISSUE #510: JUnit XML output is not well-formed
The JUnit XML output logic just puts all log output into a CDATA section without handling characters that are illegal in XML (like control characters), This will for example happen if the log output happess to be the log of a pexpect session controlling an interactive application with keyboard shoutcuts.

kwargs:
1. Escape 'invalid-xml-characters' with their U+0000 form
2. Escape `]]>` with `]]&gt;`

### DIFF
--- a/behave/reporter/junit.py
+++ b/behave/reporter/junit.py
@@ -72,6 +72,8 @@ Best sources are:
 from __future__ import absolute_import
 import os.path
 import codecs
+import re
+import sys
 from xml.etree import ElementTree
 from datetime import datetime
 from behave.reporter.base import Reporter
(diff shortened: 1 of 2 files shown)


clembou:
For the record I think this is still needed - we are having to monkey patch the formatter with a change mostly taken from this PR to keep Jenkins happy.

ncoghlan:
The code here looks reasonable and has associated tests and release notes - any chance this could be picked for the next release @jenisys @jeamland ?

ncoghlan:
@bittner Is there anything else we can do to nudge this for the next release? I've been using this code to try to intercept data before it gets to behave, but that's a game of whack-a-mole that involves a lot of guessing every time Jenkins throws its utterly useless "failed to parse XML" traceback.

bittner (approved):
This contribution looks good to me. What do you think, @jenisys?

I would merge it if you don't object. Let me know in case.

PR #815, opened 2020-03-04
Remove deprecated U mode for Python 3 compatibility.

### LINKED ISSUE #813: "U" mode is no longer accepted from Python 3.9
https://github.com/behave/behave/blob/80c1e97acdd3251dc291fbb8d1d02c774d3f6d2d/behave/json_parser.py#L36

https://github.com/python/cpython/pull/16959

tirkarthi:
Fixes #813

### DIFF
--- a/behave/json_parser.py
+++ b/behave/json_parser.py
@@ -10,6 +10,7 @@ REQUIRES: Python >= 2.6 (json module is part of Python standard library)
 
 from __future__ import absolute_import
 import codecs
+import six
 from behave import model
 from behave.model_core import Status
 try:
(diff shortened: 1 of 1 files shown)


jenisys:
Sorry, works for me is not works for everyone.
Python2 is currently still supported.

tirkarthi:
I don't get your comment. This change is backwards compatible with Python 2. Python 2 is officially not supported and has reached end of life by Jan 2020.

jenisys:
But behave still supports it.
I am not breaking stuff that was working in the past for other people.

tirkarthi:
This will work on both Python 2 and 3. It won't break anything.

bittner:
@jenisys Can you explain why this change would break backwards compatibility?

@tirkarthi Can you explain why this change will _not_ break backwards compatibility?

tirkarthi:
@bittner  The flag "U" is used in Python 2 for universal newlines. In Python 3 it's made default and hence it was deprecated for removal : https://docs.python.org/3/library/functions.html#open . see also https://stackoverflow.com/questions/56791545/what-is-the-non-deprecated-version-of-open-u-mode . So "rU" will work with Python 2 and using only "r" will help Python 3.9 compatibility. I have used six to ensure the relevant flag is passed. As a note for helping modules to migrate to Python 2 this was change was reverted to be introduced in Python 3.10 easing the compatibility concerns for libraries. I have deleted the fork since the PR was closed.

Cpython issue : https://bugs.python.org/issue37330
Reversion to be introduced in Python 3.10 : https://github.com/python/cpython/commit/942f7a2dea2e95a0fa848329565c0d0288d92e47

Thanks

PR #823, opened 2020-04-05
[MINOR] Remove unused import

deepyaman:


deepyaman:
Any reason not to address this?

jenisys:
Yes, because it is an example to show you what you can normally import.
It is not how YOU personally must do it.

PR #824, opened 2020-04-05
Create .pre-commit-hooks.yaml

deepyaman:


jenisys:
Reopen, I will look into this soon.

bittner:
@deepyaman Can you please edit the description of this issue and explain what your contribution does? :arrow_up: :memo: 

This seems like a configuration file for https://pre-commit.com/, but why is this useful? The configuration doesn't seem to do anything, or not much at least. Why would we need this in this project? Who will benefit? Can you explain?

deepyaman:
@bittner This allows Behave to be used as a `pre-commit` hook. For example, if I want to make sure (some fast subset of) Behave tests pass before pushing to my repo, I could use the widely-used Python `pre-commit` package to [configure a pre-push hook](https://pre-commit.com/#pre-commit-during-push) to make sure the tests run successfully. (I assume Behave is likely not something you want to run for each commit, based on the idea that committing should be fast, so wouldn't configure it that way.)

As an example, this is also how Black has configured it: https://github.com/psf/black/blob/master/.pre-commit-hooks.yaml
An explanation as to how it's used: https://black.readthedocs.io/en/stable/version_control_integration.html

The configuration is simple but allows users to point to the repo and use the pre-commit hook out of the box, rather than having to define a custom hook for each repo. It would become another plugin `pre-commit` users would be able to use (added to https://pre-commit.com/hooks.html).

Will copy into the description above.

ggiesen:
Would love to see this get merged as well, would like to use behave with pre-commit.

ggiesen:
Any update on this? Would still love to see it merged.

PR #834, opened 2020-05-12
Allow Context to be used in step names and text

luto:
Insert the data inside context into step names and texts, so they can be used like variables from `Scenario Outlines`. Refer to the included docs for details.

If you'd like anything implemented differently, just tell me. I am happy to adapt this to your liking.

### DIFF
--- a/behave/model.py
+++ b/behave/model.py
@@ -1767,18 +1767,18 @@ class Step(BasicStatement, Replayable):
         # self.hook_failed = False
         self.reset()
 
-        flat_context = runner.context.as_flat()
+        render_params = {}
 
-        for k, v in list(flat_context.items()):
+        for k, v in runner.context.as_flat().items():
             try:
-                flat_context[k] = str(v)
+                render_params['ctx:' + k] = str(v)
             except:
-                del flat_context[k]
+                pass
 
-        self.name = ScenarioOutlineBuilder.render_template(self.name, params=flat_context)
+        self.name = ScenarioOutlineBuilder.render_template(self.name, params=render_params)
 
         if self.text:
-            self.text = ScenarioOutlineBuilder.render_template(self.text, params=flat_context)
+            self.text = ScenarioOutlineBuilder.render_template(self.text, params=render_params)
 
         match = runner.step_registry.find_match(self)
         if match is None:
--- a/behave/model.py
+++ b/behave/model.py
@@ -1767,6 +1767,19 @@ class Step(BasicStatement, Replayable):
         # self.hook_failed = False
         self.reset()
 
+        flat_context = runner.context.as_flat()
+
+        for k in list(flat_context.keys()):
+            try:
+                flat_context[k] = str(k)
+            except:
+                del flat_context[k]
+
+        self.name = ScenarioOutlineBuilder.render_template(self.name, params=flat_context)
+
+        if self.text:
+            self.text = ScenarioOutlineBuilder.render_template(self.text, params=flat_context)
+
         match = runner.step_registry.find_match(self)
         if match is None:
             runner.undefined_steps.append(self)
--- a/behave/model.py
+++ b/behave/model.py
@@ -1769,9 +1769,9 @@ class Step(BasicStatement, Replayable):
 
         flat_context = runner.context.as_flat()
 
-        for k in list(flat_context.keys()):
+        for k, v in list(flat_context.items()):
             try:
-                flat_context[k] = str(k)
+                flat_context[k] = str(v)
             except:
                 del flat_context[k]
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -433,6 +433,31 @@ The *context* variable in all cases is an instance of
 
 .. _docid.tutorial.environmental-controls:
 
+Using Context data in steps
+---------------------------
+
+Like shown above, you can use the data in "context" to pass data between steps.
+Variables stored there can also be used in a scenario like so:
+
+.. code-block:: python
+
+    @given('I request a new widget for an account via SOAP')
+    def step_impl(context):
+        client = Client("http://127.0.0.1:8000/soap/")
+        response = client.Allocate(first='Firstname', last='Lastname')
+        context.widget_id = response['id']
+
+    @then('I delete the widget with ID "{id}"')
+    def step_impl(context, id):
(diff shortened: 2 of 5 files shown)


docs/tutorial.rst
```diff
@@ -433,6 +433,31 @@ The *context* variable in all cases is an instance of
...
+    @then('I delete the widget with ID "{id}"')
+    def step_impl(context, id):
+        client = Client("http://127.0.0.1:8000/soap/")
+        client.delete(id)
+
+.. code-block:: gherkin
+
+  Scenario: request and delete widget
+    Given I request a new widget for an account via SOAP
+    When I delete the widget with ID "<widget_id>"
```
bittner, line 459: Interesting! Is this `"<syntax>"` officially supported by the Gherkin specification?

bittner: As a second thought, why would we want to mix the description of the behavior and details of the implementation? Shouldn't we better discourage such a style of programming?

docs/tutorial.rst
```diff
@@ -433,6 +433,31 @@ The *context* variable in all cases is an instance of
...
+    @given('I request a new widget for an account via SOAP')
+    def step_impl(context):
+        client = Client("http://127.0.0.1:8000/soap/")
+        response = client.Allocate(first='Firstname', last='Lastname')
+        context.widget_id = response['id']
+
+    @then('I delete the widget with ID "{id}"')
+    def step_impl(context, id):
+        client = Client("http://127.0.0.1:8000/soap/")
+        client.delete(id)
```
bittner, line 453: I understand this example is about showcasing a typical example.

Just for clarification, let me ask a few questions:

- Why would we want to use the `widget_id` in the Gherkin-language feature file? (What if the attribute name changes?)
- Wouldn't it be wiser to use the `context.widget_id` directly in the code, instead of passing it in via the `id` parameter? (Use [declarative style](https://cucumber.io/docs/bdd/better-gherkin/#consider-a-more-declarative-style)!)

I'm not trying to make "stupid questions". Maybe I'm missing something, or the example could be improved.

What do you think?

docs/tutorial.rst
luto: Our main use case is to enable highly reusable step implementations, shared among many feature files. Specifying an ID here is rather bad form, I agree!

To be honest, I had quite a bit of trouble figuring out an example. Here is how we use it:

```
Given there is an uberspace account
When I input "uberspace setup mail"
Then the directory "/home/<username>/users" does exist
```

We don't know the account name beforehand. So the `@given(u'there is an uberspace account')` step sets `context.username` to be used by the other steps.

Maybe we can work out a better example starting from this?

behave/model.py
```diff
@@ -1767,6 +1767,19 @@ def run(self, runner, quiet=False, capture=True):
         # self.hook_failed = False
         self.reset()
 
+        flat_context = runner.context.as_flat()
+
+        for k in list(flat_context.keys()):
+            try:
+                flat_context[k] = str(k)
+            except:
```
jenisys, line 1775: Using except without Exception is BAD.

docs/tutorial.rst
jenisys, line 459: NOTE: The placeholder syntax `<placeholder>` works only in a ScrenarioOutline, not in a Scenario.

jenisys (changes requested):
@luto I am not quite sure if I cannot accept the PR as it is right now.

With your proposed enhancement, there are now two sources for placeholders 
(actually: 3 by including the internally provided placeholders):
* Column names of column headers in Examples table(s)
* Context variables

QUESTIONS:
* What happens if the same name exists in both placeholder name providers ?!?
* What happens if the referenced context variable does not exist ?

It would be cleaner to distinguish between both, like using `<ctx:context_variable_name>` or `<context:context_variable_name>` to specify that the placeholder should be retrieved from the context stack.

NOTES:
This is the "fast feedback" after looking over your proposed changes just shortly (in addition to @bittner 's points).
I still need to look into your changes in more detail. Therefore, more may come up.

behave/model.py
luto: true! I'll fix this once we agree on a goal and syntax.

docs/tutorial.rst
luto: Part of the point of this PR is to enable them in `Scenario`s as well.

luto:
> I am not quite sure if I can accept the PR as it is right now.

No worries! I am happy to rework this, so it meets your standards :)

> What happens if the referenced context variable does not exist ?

I'd say we error and abort in that case. What do you think?

> It would be cleaner to distinguish between both, like using <ctx:context_variable_name> or <context:context_variable_name> to specify that the placeholder should be retrieved from the context stack.

Fine by me! @bittner does that work for you too?

bittner:
> Fine by me! @bittner does that work for you too?

I would prefer if you replied to my comments directly. Jens' comments are Jens', mine are mine.

docs/tutorial.rst
jenisys: The "normal way" to avoid this is to use a "persona" oriented approach (or user-roles/host-roles and instance numbers, if possible/sensible).You basically use names to identify configuration dataset that you want to use.
The dataset can be provided by using the userdata mechanism (or similar other mechanisms). Within the step you just need to identify which dataset to use and the step specifies what should be done (note: The identification of the dataset may occur in an earlier step, often a given-step).

HINT:
The concrete IPv4 address may be unnecessary technical details in BDD steps (a smell that hints that you may provide too many low-level details which may clutter your steps and hide what is really important in the steps).

RELATED: Personas
* https://www.interaction-design.org/literature/article/personas-why-and-how-you-should-use-them
* https://www.testingtime.com/en/blog/complete-guide-how-to-create-personas-based-on-data/

jenisys: > Interesting! Is this `"<syntax>"` officially supported by the Gherkin specification?

It is the placeholder syntax in ScenarioOutline templates.
That problem is that newer Gherkin version want to make Scenario and ScenarioOutline interchangeable.
Therefore, you can use the keyword `Scenario` but it is actually a `ScenarioOutline` / `ScenarioTemplate`.
This suggested placeholder syntax above will prevent to support this Gherkin functionality in `behave` in the future (same syntax used for two different things).

jenisys:
DECISION: Will not be merged

PR #840, opened 2020-06-02
Update install.rst

fredericLeguedois:
Add Ubuntu 20.04 command line to install Behave

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -22,13 +22,23 @@ To update an already installed :pypi:`behave` version, use:
 
 .. _`pip related information`:  https://pip.pypa.io/en/latest/installing/
 
-Using Ubuntu 20.04
-------------------
-
-sudo apt install python3-behave
 
 Using a Source Distribution
 ---------------------------
+
+* `Overview`_
+
+.. _`Overview`:  http://rpm.pbone.net/info_idpl_70083425_distro_fedora_32_com_python3-behave-1.2.6-6.fc32.noarch.rpm.html
+
+* `Ubuntu`_
+
+.. _`Ubuntu`: https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-behave&searchon=names
+
+    sudo apt install python3-behave
+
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -22,6 +22,10 @@ To update an already installed :pypi:`behave` version, use:
 
 .. _`pip related information`:  https://pip.pypa.io/en/latest/installing/
 
+Using Ubuntu 20.04
+------------------
+
+sudo apt install python3-behave
```
bittner, line 28: I'd rather add a section on "Using Your Package Manager" and list several options. We'll end up either covering only a very small part of the target population otherwise. And the information will out-date quickly anyway if we mention just a single release.

Examples:
- [Overview](https://repology.org/project/python:behave/versions)
- [Ubuntu](https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-behave&searchon=names) (your example above, valid for several releases)
- [Fedora](http://rpm.pbone.net/info_idpl_70083425_distro_fedora_32_com_python3-behave-1.2.6-6.fc32.noarch.rpm.html)

Would you like to update your change to be more generic?

bittner (changes requested):

docs/install.rst
jenisys: @fredericLeguedois Sorry to say, but I see it the same way like @bittner .

fredericLeguedois: I did the change.

bittner: Thanks Frédéric!

The result doesn't read logic, though.

I'm trying to not be too picky, but first the formatting is a bit odd (three separate bullets instead of a neat list), second the links don't seem to fit into the section "Source Distribution". Do you want to make this a bit more logical?

Note that my examples list from above is really just a collection of links I found, not a final content suggestion.

fredericLeguedois: Thanks Petter for your feedback. I don't understand what to do. Maybe it's better to cancel the commit.

PR #844, opened 2020-06-26
Docs: change code blocks from bash to console

ffe4:
Changed code block language of interactive shell session examples from `bash` to `console`. 

`bash` is used for shell scripts, rather than interactive sessions, and highlights certain words in the console output on both GitHub and PyPI. Changing the code block language to `console` should work on both PyPI (Pygments) and GitHub.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -76,7 +76,7 @@ In that directory create a file called "example_steps.py" containing:
 
 Run behave:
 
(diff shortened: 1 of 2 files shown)


bittner (approved):

bittner:
Thanks Daniel! :+1:

PR #845, opened 2020-06-29
Enhancement: repeated test runs (loop)

bbouv:
Add new loop configuration parameter to run tests more than once.
This allows to have only one report for the same tests running
several times.

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -95,6 +95,10 @@ options = [
      dict(metavar="PATTERN", dest="include_re",
           help="Only run feature files matching regular expression PATTERN.")),
 
+    (("-l", "--loop"),
+     dict(action="store", default=1, type=int,
+          help="Define loop number to execute tests")),
+
     (("--no-junit",),
      dict(action="store_false", dest="junit",
           help="Don't output JUnit-compatible reports.")),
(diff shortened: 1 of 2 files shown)


CsatariGergely:
@jenisys , @jeamland or @r1chardj0n3s can you please review?

jenisys:
This approach is too simplistic.
You should get the wrong results with reporters (SummaryReporter, JUnitReporter), basically only the results of the last loop iteration. The list of features is handed to the reporters at the end of the test run.

Therefore, for this functionality, the feature objects need to be copied N-1 times and appended to the features list before the test run.

bittner (changes requested):
@jenisys Can you give a concrete example, a code snippet or a reference thereto, on how the copying should be attempted? To incentivise @bbouv to improve the contribution into the right direction.

bittner:
This contribution would also need to extend our documentation.

jenisys:
NOTE: The name/option `--repeat=NUMBER` is more commonly used (at least for me).
It basically provides the functionality of a repeated test runner.

jenisys:
CRITICAL POINTS ARE (or need to be looked over/addressed):
* Test run infrastructure support to duplicate/clone existing model elements (Features)
  REASON: Scenarios may fail in one test run, but not the other (flaky tests).
  Therefore, an entity that provides the test run results for each test run is needed.
* JUnit XML test reports
* JSON formatter output

PR #877, opened 2020-11-15
API reference - Capitalizing Steps on Steps function

Ibrian93:
A minor change where I capitalize the steps on the section **Step Function**
That causes that the current editor highlight the words properly

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -74,10 +74,10 @@ the name of their preceding keyword, so given the following feature file:
 .. code-block:: gherkin
 
     Given some known state
-      and some other known state
-     when some action is taken
-     then some outcome is observed
-      but some other outcome is not observed.
+      And some other known state
+     When some action is taken
+     Then some outcome is observed
+      But some other outcome is not observed.
 
 the first "and" step will be renamed internally to "given" and *behave*
 will look for a step implementation decorated with either "given" or "step":


bittner (approved):
This change matches the casing used in the [official Gherkin specification](https://cucumber.io/docs/gherkin/reference/).

On the other hand, the specification says:

> The recommended indentation level is two spaces.

We would probably need to go over our entire docs and iron this out, separately.

jenisys:
NOTES:
* Normal convention within behave document currently is using Uppercase keywords (IMHO).
  Therefore, this PR goes in this direction.
* Behave supports Keywords starting with Uppercase letter or lowercase.
  Internally, only lowercase words are used to look up the language specific keywords.
* Only because your IDE/Editor suggests does not mean it is correct. It is a tool, and tools may be correct or may be wrong.

SEE ALSO:
* [behave/i18n.py](https://github.com/behave/behave/blob/master/behave/i18n.py#L23)

jenisys:
INTEGRATED INTO: behave-1.2.7.dev2

jenisys:
@bittner Regarding whitespace changed (2 spaces or 4 spaces).
I would suggest we resolve that internally and provide a convention (and/or cleanup the docs afterwards if needed).
Currently, the convention is specified by ".editorconfig" file in the root of the repository: 2 space for `*.feature` files

PR #887, opened 2020-12-12
Add Contributing chapter

bittner:
Adds a new chapter aimed at giving guidance on contributing to Behave.

## Tox configuration

Also adapts the `docs` Tox target, so that it generates the docs at the same location than `invoke docs` (not inside `.tox/tmp/`). This gives us a unified user experience and, as a nice side-effect, the `invoke docs.clean` command will clean up "both" locations.

## Issues with Requirements

Having Tox install `behave` when `tox -e docs` is run requires DocUtils to be prior installed due to some setup that makes `setuptools` verify the package README. We should try to remove this dependency and have the proper README generation checked independently, e.g. [using Twine](https://github.com/bittner/pyclean/blob/master/tox.ini#L46-L51) via a Tox target.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -108,7 +108,7 @@ if USE_SPHINX_INTERNATIONAL:
 # -----------------------------------------------------------------------------
 project = u"behave"
 authors = u"Jens Engel, Benno Rice and Richard Jones"
-copyright = u"2012-2019, %s" % authors
+copyright = u"2012-2020, %s" % authors
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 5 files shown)


bittner:
@jenisys I configured a [web hook](https://github.com/behave/behave/settings/hooks) for ReadTheDocs to [build the documentation](https://readthedocs.org/dashboard/behave/integrations/) automatically. I will merge this now, so we can verify whether this works as expected. If there are any corrections on the content of this PR let's fix them with a follow-up PR.

bittner:
Works! :rocket: The "latest" docs are building. :gear: :heavy_check_mark:

bittner:
[Contributing chapter](https://behave.readthedocs.io/en/latest/contributing.html) is now available in the docs. :book:

PR #894, opened 2021-01-09
FIX: Some regressions in test suite for python 3.9

jenisys:
Fix some regressions that occur when using Python 3.9:

* async-steps: Use more stable active-tags now.
* JUnit XML related: Adapt to XML attribute ordering changes since Python 3.8 (add missing active-tag for python=3.9).
* Prepare active-tags usage for Python 3.10

OTHERWISE:

* Behave jsonschema: Add some extra elements (related to: gherkin v6 Rule).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ tools/virtualenvs
 .ropeproject
 nosetests.xml
 rerun.txt
+testrun*.json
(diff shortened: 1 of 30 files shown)


jenisys:
INTEGRATED INTO: behave 1.2.7.dev2

PR #923, opened 2021-05-15
Feature/test retry delay and backoff

connor-philip:


behave/contrib/scenario_autoretry.py
```diff
@@ -48,17 +49,28 @@ def patch_scenario_with_autoretry(scenario, max_attempts=3):
...
                 if attempt > 1:
                     message = u"AUTO-RETRY SCENARIO PASSED (after {0} attempts)"
                     print(message.format(attempt))
                 return False    # -- NOT-FAILED = PASSED
+
             # -- SCENARIO FAILED:
             if attempt < max_attempts:
-                print(u"AUTO-RETRY SCENARIO (attempt {0})".format(attempt))
+                if delay:
+                    if attempt > 1: _delay *= backoff
```
bittner, line 68: Please move the assignment to the next line. Single-line `if`s are known as bad style.

behave/contrib/scenario_autoretry.py
```diff
@@ -48,17 +49,28 @@ def patch_scenario_with_autoretry(scenario, max_attempts=3):
...
+                    delay_message =  u" Delaying retry for {0}s".format(_delay)
+
+                print(u"AUTO-RETRY SCENARIO (attempt {0}){1}".format(attempt, delay_message))
```
bittner, line 71: This is a bit hacky. A user will be required to start the `delay_message` with a blank character to make the output display decently.

connor-philip: Can you elaborate? The user wouldn't be setting the `delay_message` here, it's added onto the current message which is printed. Nothing extra will be printed if the user hasn't added a delay.

bittner: I understand. I wonder if there was a more elegant way to code this. :thinking:

features/runner.scenario_autoretry.feature
```diff
@@ -128,4 +134,123 @@ Feature: Auto-retry failed scenarios (a number of times)
...
+      
+        Scenario: A2
+          Given an unreliable step fails sometimes ... failed
+        Assertion Failed: UNRELIABLE-STEP FAILURE
+        AUTO-RETRY SCENARIO (attempt 1) Delaying retry for 1s
+      
+        Scenario: A2
+          Given an unreliable step fails sometimes ... passed
+          Then another step passes ... passed
+        """
```
bittner, line 256: Can you add a final newline here, please?

bittner (approved):
@jenisys Please take a look at this PR. Seems like an improvement worth merging.

EnricoSemrau (approved):
LGTM 👍🏻

PR #925, opened 2021-05-21
Fix junit JUnit XML output. Fixes #510

### LINKED ISSUE #510: JUnit XML output is not well-formed

lawnmowerlatte:
1. Escape 'invalid-xml-characters' with their U+0000 form
2. Escape ']]>' with ']]&gt;'

### DIFF
--- a/behave/reporter/junit.py
+++ b/behave/reporter/junit.py
@@ -72,6 +72,8 @@ Best sources are:
 from __future__ import absolute_import
 import os.path
 import codecs
+import re
+import sys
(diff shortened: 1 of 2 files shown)


lawnmowerlatte:
Sorry, was intended to PR against a fork.

Issue #641, opened 2018-03-27
Pylint errors when importing `given - when - then` from `behave`


lktslionel:
Hi,

I'm have some issues with the import statement : 
```
from behave import given, when, then
```

I use `pylint-v1.8` and it shows the following errors in my editor (vscode) : 

```
[pylint] E0611:No name 'given' in module 'behave'
[pylint] E0611:No name 'when' in module 'behave'
[pylint] E0611:No name 'then' in module 'behave'
```

Below some info about my environment : 

```
Python: 3.6.4
pipenv, version 11.8.3
behave 1.2.6
```

- Any idea of why pylint is showing errors ? 
- Why It can't find `behave` annotations ? 

Thanks

jenisys:
Pylint seems to have a problem w/ wildcard-imports that are perfectly legal in Python. 
If you replace the wildcard-import with explicit imports, everything is fine for pylon.
Add an issue to pylint to get this fixed.

rrueth:
I ran across this issue while starting to adopt behave today. @jenisys, what do you mean by "if you replace the wildcard-import with explicit imports, everything is fine?"  If I do:

```
import behave

@behave.when("I run a test")
def foo():
  pass
```

I get the following error from `pylint`: 
```
[E1101(no-member), create_user] Module 'behave' has no 'given' member
```

Is there a way that I can import `behave` and `given`/`when`/`then` that would eliminate the error?

For reference, these are the two issues that I found on the pylint side:
- https://github.com/PyCQA/pylint/issues/535
- https://github.com/PyCQA/pylint/issues/1394

jenisys:
@rrueth 
The core question is if the python-interpreter is right or pylint is right.
The used construct in ``behave`` is a completely legal construct in python AFAIK (and older versions of pylint did not complain about it).

SEE:
* https://www.python.org/dev/peps/pep-0008/#id24

NOTE: By the way, why do you get an error on "given" when you are not using it ?!?

rrueth:
Sorry, my mistake in the example.  I copied the wrong line of my pylint ouput =).  It is actually:

```
[E1101(no-member), create_user] Module 'behave' has no 'when' member
``` 

The same error would show up for `given` and `then` if I had done `@behave.given` or `@behave.then` instead.

[...]

PR #967, opened 2021-10-14
Update __init__.py in behave import to fix pylint

dsayling:
#641 complains about having to disable the no-name-in-module import error everywhere behave is imported OR you have to disable wildcard-import 
either way - users using strict configurations of pylint are burdened with this

to simply get around it, if the steps are directly imported in `behave/__init__.py` this is no longer a burden on the user - if for some reason a new step type is added to the registry, this line would need to be updated (but so would the `__all__`), so I dont think this is too much of a burden on improvements to behave

### DIFF
--- a/behave/__init__.py
+++ b/behave/__init__.py
@@ -17,7 +17,7 @@ To get started, we recommend the `tutorial`_ and then the `test language`_ and
 """
 
 from __future__ import absolute_import
-from behave.step_registry import *      # pylint: disable=wildcard-import
+from behave.step_registry import given, when, then, step, Given, When, Then, Step      # pylint: disable=no-name-in-module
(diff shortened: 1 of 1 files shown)


jenisys:
@dsayling 
Nice catch, I will look into it (and merge it soon).

NOTE:
It is just weird for me why `pylint` does not cover that (even when `from xxx import *` is normally not advised).

dsayling:
@jenisys Thanks! And somewhat annoyingly, pylint has been ignoring `__all__` for sometime now. [See this issue](https://github.com/PyCQA/pylint/issues/2392#issuecomment-411645504).

jenisys:
INTEGRATED INTO: behave 1.2.7

Issue #985, opened 2021-12-27
Existing loggers are overwritten by file config


janoskut:
When we use `context.config.setup_logging(configfile=...)`, existing loggers are always overwritten by the new configuration. This is the default behavior by the `logging` module, and noted with a big warning in https://docs.python.org/3/howto/logging.html#configuring-logging:

> Warning: The fileConfig() function takes a default parameter, disable_existing_loggers, which defaults to True for reasons of backward compatibility. This may or may not be what you want, since it will cause any non-root loggers existing before the fileConfig() call to be disabled unless they (or an ancestor) are explicitly named in the configuration. Please refer to the reference documentation for more information, and specify False for this parameter if you wish.

In `behave`, the `before_all()` hook is normally invoked only after all other modules are imported, and hence *after* all file-global loggers are created. While it may be debatable, it is common practice to use file-global loggers such as:

```py
import logging

log = logging.getLogger(__name__)

def foo():
    log.info('using the global logger')
```

When using such modules in `behave` with file-config, those global loggers are overwritten, and all modules which use such loggers (e.g. step definitions, or from the system under test) won't provide logging.

I suggest to at least pass all `**kwargs` to `logging.config.fileConfig()`, to allow to overwrite this default behavior, or even more change the default behavior to "not disable existing loggers" (and still pass the kwargs to allow control of it).

See the https://github.com/behave/behave/pull/986.

jenisys:
SHOULD BE RESOLVED BY: behave v1.2.7.dev7 (or current HEAD of the repository)

* Improved logging support

SEE: 
* 
* https://behave.readthedocs.io/en/latest/new_and_noteworthy_v1.2.7/#improved-logging-support

PR #986, opened 2021-12-27
#985 Don't disable existing loggers by default

janoskut:
Rationale:
In behave, loggers are usually initialized before "before_all()" and "setup_logging()" are invoked.
When using a file config, the default behavior of the logging module is to disable existing loggers (see warning in https://docs.python.org/3/howto/logging.html#configuring-logging).
Changing the default behavior to not disabling the existing loggers matches behaves logger users better.
By passing the **kwargs into "fileConfig()", we still allow all desired behaviors.


jenisys:
Correction of description above:
* `behave` does not initialize the logging subsystem before calling the `before_all()` hook. If this is the case, it will be corrected.
* BUT: The user code that is imported via „environment.py“ may pre-initialize the logging subsystem.
* The `logging` module/subsystem uses the init-once only strategy which causes must of the problems or misunderstandings when using it.

janoskut:
@jenisys  this actually exactly what I meant. Does it make it clearer if I change the first sentence to:

>  In behave, loggers are usually initialized **by user code** before "before_all()" and "setup_logging()" are invoked.

More rationale in #985.

jenisys:
No, I just wanted to clarify the point that it can be understood by others when reading the ticket.

janoskut:
@jenisys  is there something I can do to push this change further, like help to clarify any doubts, or add more tests?

jenisys:
I am working on it.

jenisys:
Should be resolved by improved logging support (how to set up the logging subsystem).

Issue #848, opened 2020-07-14
Additional external documentation/presentation links


ncoghlan:
I noticed https://behave.readthedocs.io/en/latest/more_info.html today, and wanted to offer couple more links that may be of interest:

* Beginning BDD with Django (Nicole Harris):
    * https://whoisnicoleharris.com/2015/03/16/bdd-part-one.html
    * https://whoisnicoleharris.com/2015/03/19/bdd-part-two.html
* Front-end integration testing with splinter (and behave!):
    * https://pyvideo.org/pycon-au-2017/front-end-integration-testing-with-splinter.html

PR #989, opened 2022-01-09
Add link to tutorial by Nicole Harris and talk by Nick Coghlan

bittner:
As suggested by @ncoghlan via #848.


jenisys (approved):
I assume „BDD with Python“ is outdated?
Otherwise OK.

PR #991, opened 2022-01-10
feat(config): parse pyproject.toml

### LINKED ISSUE #944: pyproject.toml (PEP 518) support
Hi! Is there a plan to support `pyproject.toml` from [PEP 518](https://www.python.org/dev/peps/pep-0518/)?

Specifically, it would be nice if `behave` could understand `[tool.behave]` section in the `pyproject.toml` file like other python libraries do (e.g. `black`, `coverage`, `isort`, `mypy`, `pytest`, ...)

pwoolvett:
fixes #944
See related discussion

### DIFF
--- a/behave.ini
+++ b/behave.ini
@@ -1,7 +1,7 @@
 # =============================================================================
 # BEHAVE CONFIGURATION
 # =============================================================================
-# FILE: .behaverc, behave.ini, setup.cfg, tox.ini
+# FILE: .behaverc, behave.ini, setup.cfg, tox.ini, pyproject.toml
 #
 # SEE ALSO:
 #  * http://packages.python.org/behave/behave.html#configuration-files
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -3,6 +3,7 @@
 from __future__ import absolute_import, print_function
 import argparse
 import inspect
+import json
 import logging
 import os
 import re
@@ -27,6 +28,17 @@ from behave.textutil import select_best_encoding, to_texts
 ConfigParser = configparser.ConfigParser
 if six.PY2:
     ConfigParser = configparser.SafeConfigParser
+
+try:
+    if sys.version_info >= (3, 11):
+        import tomllib
+    elif sys.version_info < (3, 0):
+        import toml as tomllib
+    else:
+        import tomli as tomllib
+    _TOML_AVAILABLE = True
+except ImportError:
+    _TOML_AVAILABLE = False
 
 
 # -----------------------------------------------------------------------------
@@ -382,41 +394,49 @@ raw_value_options = frozenset([
 ])
 
 
-def read_configuration(path):
-    # pylint: disable=too-many-locals, too-many-branches
-    config = ConfigParser()
-    config.optionxform = str    # -- SUPPORT: case-sensitive keys
-    config.read(path)
-    config_dir = os.path.dirname(path)
-    result = {}
+def values_to_str(d):
+    return json.loads(
+        json.dumps(d),
+        parse_float=str,
+        parse_int=str,
+        parse_constant=str
+    )
+
+
+def decode_options(config):
     for fixed, keywords in options:
         if "dest" in keywords:
             dest = keywords["dest"]
         else:
+            dest = None
             for opt in fixed:
                 if opt.startswith("--"):
                     dest = opt[2:].replace("-", "_")
                 else:
                     assert len(opt) == 2
                     dest = opt[1:]
-        if dest in "tags_help lang_list lang_help version".split():
+        if (
+            not dest
+        ) or (
+            dest in "tags_help lang_list lang_help version".split()
+        ):
             continue
-        if not config.has_option("behave", dest):
+        try:
+            if dest not in config["behave"]:
+                continue
+        except AttributeError as exc:
+            # SafeConfigParser instance has no attribute '__getitem__' (py27)
+            if "__getitem__" not in str(exc):
+                raise
+            if not config.has_option("behave", dest):
+                continue
+        except KeyError:
             continue
         action = keywords.get("action", "store")
-        if action == "store":
-            use_raw_value = dest in raw_value_options
-            result[dest] = config.get("behave", dest, raw=use_raw_value)
-        elif action in ("store_true", "store_false"):
-            result[dest] = config.getboolean("behave", dest)
-        elif action == "append":
-            if dest == "userdata_defines":
-                continue    # -- SKIP-CONFIGFILE: Command-line only option.
-            result[dest] = \
-                [s.strip() for s in config.get("behave", dest).splitlines()]
-        else:
-            raise ValueError('action "%s" not implemented' % action)
+        yield dest, action
 
+
+def format_outfiles_coupling(result, config_dir):
     # -- STEP: format/outfiles coupling
     if "format" in result:
         # -- OPTIONS: format/outfiles are coupled in configuration file.
@@ -442,6 +462,32 @@ def read_configuration(path):
             result[paths_name] = \
                 [os.path.normpath(os.path.join(config_dir, p)) for p in paths]
 
+
+def read_configparser(path):
+    # pylint: disable=too-many-locals, too-many-branches
+    config = ConfigParser()
+    config.optionxform = str    # -- SUPPORT: case-sensitive keys
+    config.read(path)
+    config_dir = os.path.dirname(path)
+    result = {}
+
+    for dest, action in decode_options(config):
+        if action == "store":
+            result[dest] = config.get(
+                "behave", dest, raw=dest in raw_value_options
+            )
+        elif action in ("store_true", "store_false"):
+            result[dest] = config.getboolean("behave", dest)
+        elif action == "append":
+            if dest == "userdata_defines":
+                continue    # -- SKIP-CONFIGFILE: Command-line only option.
+            result[dest] = \
+                [s.strip() for s in config.get("behave", dest).splitlines()]
+        else:
+            raise ValueError('action "%s" not implemented' % action)
+
+    format_outfiles_coupling(result, config_dir)
+
     # -- STEP: Special additional configuration sections.
     # SCHEMA: config_section: data_name
     special_config_section_map = {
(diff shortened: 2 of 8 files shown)


pwoolvett:
@jenisys note: current implementation does not behave (🤣 ) as your modification in  the comment from #944.

Specifically, when the cli option has`action=='append'`,  i'm requiring an array from toml instrad of  the '.splitlines'

the not so obvious implication is that for it to work, it *has* to be an array, which migh be counterintuitive:

```toml
# does not work in current implementation
[tool.behave]
default_tags = "foo"
```

```toml
# works in current implementation
[tool.behave]
default_tags = ["foo"]
```

bittner:
Can you rebase your changes and then run `flake8` and `pylint` over the modules you have changed?

I believe, the tools will give you a few hints on formatting and such. If that reports too much you should only worry about the code parts you have changed.

bittner:
> the not so obvious implication is that for it to work, it _has_ to be an array, which migh be counterintuitive:

@jenisys Maybe that's fine, though. In the end, the option is called in its plural form ("tags"). For me, this automatically triggers "list", also because TOML is more strict with formatting.

jenisys:
Just a comment, related to the example with the `default_tags`.
The array/list will be replaced with a string because we will use tag-expressions v2 (that contain a logical Boolean expression as string). In addition, tag-expressions v2 allow you to specify in a more concise, easier to read and understand what you want to select.

OTHERWISE: Config Data Model if option/param is a array/list
If a CLI option is a list/array, you need to decide what is meant:
* Do you want to override the existing value(s)
* Do you want to append to the the list of existing values (from a config-file)

One way to avoid this problem is to have actually 2 options/params:
* One for the default value(s) if none are provided on CLI (override mode)
* One for the value that may also be used in the config-file (append mode)

jenisys:
@bittner The name `--tags/tags/default_tags` is for historical reasons because you needed multiple terms for a real `tag-expression`.
The real, underlying name would be `tag-expression` (singular) that is used to select elements with these tags.
The current support for a list in the config-file was for backward compatibility.
This will be dropped in the future.

behave/configuration.py
```diff
@@ -28,6 +28,12 @@
 if six.PY2:
     ConfigParser = configparser.SafeConfigParser
 
+try:
+    import toml
+    _TOML_AVAILABLE = True
+    import json
+except ImportError:
+    _TOML_AVAILABLE = False
```
pauloxnet, line 36: `json` is int he standard library I think we can import it before of the `try` 
```suggestion
import json
try:
    import toml
    _TOML_AVAILABLE = True
except ImportError:
    _TOML_AVAILABLE = False
```

pauloxnet (changes requested):

pwoolvett:
> Can you rebase your changes and then run `flake8` and `pylint` over the modules you have changed?

Done, done, and done

> Just a comment, related to the example with the `default_tags`. The array/list will be replaced with a string because we will use tag-expressions v2 (that contain a logical Boolean expression as string). In addition, tag-expressions v2 allow you to specify in a more concise, easier to read and understand what you want to select.

What? how? I like the sound of that...

> 
> OTHERWISE: Config Data Model if option/param is a array/list If a CLI option is a list/array, you need to decide what is meant:
> 
>     * Do you want to override the existing value(s)
> 
>     * Do you want to append to the the list of existing values (from a config-file)
> 
> 
> One way to avoid this problem is to have actually 2 options/params:
> 
>     * One for the default value(s) if none are provided on CLI (override mode)
> 
>     * One for the value that may also be used in the config-file (append mode)

I think this last alternative sounds good, eg using `tags` for config and `extra-tags` for appending form cli. Otherwise, `tags` form cli overrides... but I'm not sure how this affects this PR yet

pauloxnet (approved):

bittner:
@jenisys Can we merge this contribution? It would be a valuable change.

behave.ini
```diff
@@ -1,7 +1,7 @@
 # =============================================================================
 # BEHAVE CONFIGURATION
 # =============================================================================
-# FILE: .behaverc, behave.ini, setup.cfg, tox.ini
+# FILE: .behaverc, behave.ini, setup.cfg, tox.ini, pyproject.toml
```
pauloxnet, line 4: Maybe we can preserve the alphabetical order
```suggestion
# FILE: .behaverc, behave.ini, pyproject.toml, setup.cfg, tox.ini
```

behave/configuration.py
```diff
@@ -452,14 +487,80 @@ def read_configuration(path):
     return result
 
 
+def read_toml(path):
+    """Read configuration from pyproject.toml file.
```
pauloxnet, line 491: I think first line comment one line below is more compliant.
```suggestion
    """
    Read configuration from pyproject.toml file.
```

behave/configuration.py
```diff
@@ -452,14 +487,80 @@ def read_configuration(path):
...
 def config_filenames():
     paths = ["./", os.path.expanduser("~")]
     if sys.platform in ("cygwin", "win32") and "APPDATA" in os.environ:
         paths.append(os.path.join(os.environ["APPDATA"]))
 
     for path in reversed(paths):
-        for filename in reversed(
-                ("behave.ini", ".behaverc", "setup.cfg", "tox.ini")):
+        for filename in reversed((
+            "behave.ini", ".behaverc", "setup.cfg", "tox.ini", "pyproject.toml"
```
pauloxnet, line 562: Maybe we can preserve the alphabetical order
```suggestion
            "behave.ini", ".behaverc", "pyproject.toml", "setup.cfg", "tox.ini"
```

docs/behave.rst
```diff
@@ -287,9 +286,9 @@ You can also exclude several tags::
 Configuration Files
 ===================
 
-Configuration files for *behave* are called either ".behaverc",
-"behave.ini", "setup.cfg" or "tox.ini" (your preference) and are located in
-one of three places:
+Configuration files for *behave* are called either ".behaverc", "behave.ini",
+"setup.cfg", "tox.ini", or "pyproject.toml" (your preference) and are located
+in one of three places:
```
pauloxnet, line 291: Maybe we can preserve the alphabetical order
```suggestion
Configuration files for *behave* are called either ".behaverc", "behave.ini",
"pyproject.toml", "setup.cfg" or "tox.ini" (your preference) and are located
in one of three places:
```

docs/behave.rst-template
```diff
@@ -29,9 +29,9 @@ Tag Expression
 Configuration Files
 ===================
 
-Configuration files for *behave* are called either ".behaverc",
-"behave.ini", "setup.cfg" or "tox.ini" (your preference) and are located in
-one of three places:
+Configuration files for *behave* are called either ".behaverc", "behave.ini",
+"setup.cfg", "tox.ini", or "pyproject.toml" (your preference) and are located
+in one of three places:
```
pauloxnet, line 34: Maybe we can preserve the alphabetical order
```suggestion
Configuration files for *behave* are called either ".behaverc", "behave.ini",
"pyproject.toml", "setup.cfg" or "tox.ini" (your preference) and are located
in one of three places:
```

pauloxnet (changes requested):
I've reviewed again this PR and it look ready to be merged.
I only added some very minor code change suggestions.

behave/configuration.py
bittner: Paolo, I believe adding the new configuration file to the end of the list we follow the principle of least surprise. Using alphabetical order just for the sake of aesthetics could have a dramatic impact. :thinking: 

**Example:** A developer that had added some config to `pyproject.toml` (just for fun, w/o having it working), having their productive setup in `setup.cfg` or `tox.ini`, may get a broken setup all of a sudden for no apparent reason.

pauloxnet: It makes sense to me. I hope you can merge this one now. 🙂

pauloxnet:
I think this PR is ready to be merged after quite a year of waiting. There's something to be improved that block the merge ?

tests/unit/test_configuration.py
```diff
@@ -44,15 +61,18 @@ class CustomTestRunner(BaseRunner):
...
-        with open(tn, "w") as f:
-            f.write(TEST_CONFIG)
-
+    @pytest.mark.parametrize(
+        ("filename", "contents"),
+        list(TEST_CONFIGS)
+    )
+    def test_read_file(self, filename, contents):
+        tndir = tempfile.mkdtemp()
+        file_path = tndir + "/" + filename
```
bittner, line 70: This will not be portable, I fear. Please use `os.path.normpath(os.path.join(...))`. See below.

bittner (changes requested):

bittner:
@pwoolvett Can you please rebase the PR? Also, we may want to mention in the installation docs that one has to install the "toml" package for TOML support, e.g. via `pip install behave[toml]`.

@jenisys Other than that, are you fine to merge this PR? This way we would get feedback on TOML being used by everyone installing off HEAD. I guess that would be okay to do.

behave/configuration.py
```diff
@@ -452,14 +487,80 @@ def read_configuration(path):
...
+        )):
```
bittner, line 563: Why do we have `reversed` here? Doesn't this reverse the preference of the configuration files, i.e. make "pyproject.toml" be found & read first, and "behave.ini" only last? If yes, we probably want to place "pyproject.toml" on the top of the list to ensure we stay backward-compatible?

Also, why don't we simply denote the order as it is? e.g. 

```suggestion
        for filename in ["tox.ini", "setup.cfg", ".behaverc", "behave.ini", "pyproject.toml"]:
```

Side note: Maybe academic, but I'd prefer to have a list here instead of a tuple. Times have changed, and so have ideas about runtime optimization. Items of an identical type should be kept in a list, differing types in a tuple.

pwoolvett:
rebased, corrected one suggestion, replied another. The ball is in your court for the moment @bittner :)

pauloxnet:
For Python 3.11 are we using the toml module in the standard library?

pwoolvett:
replaced `toml` with `tomllib`/`tomli`

where should that additional line be (`"tomli>=1.1.0; python_full_version < '3.11.0a7'"`)?
for the moment, it's just in `setup.py`, but i think it should be addded as well in `py.requirements/basic.txt` or `py.requirements/testing.txt`. Is there a script to do it or should I just add a line manually there?

bittner:
There is a failing test for Python 2.7. The problem seems to be the bracket-based syntax, which is appears to be invalid for Python < 3.3 [according to StackOverflow](https://stackoverflow.com/a/16407465/202834).

@pwoolvett  Can you fix that, so we get ready for merging? :wrench: 

@jenisys I would merge this PR afterwards, if you don't object. :rocket: 
Do you want to take a look yourself? :eyes:

setup.py
```diff
@@ -135,6 +135,10 @@ def find_packages_by_root_package(where):
         'formatters': [
             "behave-html-formatter",
         ],
+        'toml': [ # Enable pyproject.toml support.
+            "tomli>=1.1.0; python_version >=  '3.0' and python_full_version < '3.11.0a7'",
+            "toml>=0.10.2; python_version <  '3.0'", # py27 support
```
bittner, line 140: Flake8 wants us to place 2 spaces before comments.

What if we have Python >= 3.11.0a7? Neither tomli nor toml will be installed, will they? Do we have a solution for the user?

behave/configuration.py
```diff
@@ -452,14 +497,82 @@ def read_configuration(path):
...
 
+def read_toml(path):
+    """Read configuration from pyproject.toml file.
+
+    Configuration should be stored inside the 'tool.behave' table.
+
+    See https://www.python.org/dev/peps/pep-0518/#tool-table
+    """
+    # pylint: disable=too-many-locals, too-many-branches
+    with open(path, "rb") as f:
```
bittner, line 508: Why `"rb"`? Shouldn't we use read in text mode for cross-platform compatibility?

```suggestion
    with open(path) as tomlfile:
```

We shouldn't trust that configuration readers can handle Windows and Unix newlines interchangeably.

Pylint suggests to use self-explanatory variable names.

py.requirements/testing.txt
```diff
@@ -20,5 +20,9 @@ PyHamcrest <  2.0;   python_version <  '3.0'
 # HINT: path.py => path (python-install-package was renamed for python3)
 path.py >=11.5.0,<13.0; python_version <  '3.5'
 path >= 13.1.0;         python_version >= '3.5'
+# NOTE: toml-0.10.2 is required for python<3 (tomli not available)
+tomli>=1.1.0; python_version >=  '3.0' and python_full_version < '3.11.0a7'
+toml>=0.10.2; python_version <  '3.0'
+
```
bittner, line 26: It might be possible to refer to setup.py's `[toml]` extras like this:

```suggestion
.[toml]
```

Do you want to try?

bittner (commented):
Could you please also address the remaining review comments? Thank you!

pauloxnet:
> There is a failing test for Python 2.7.

Behave is still supporting Python2 after 2 years of it's EOL date ?

I think it's time to remove support for Python2, I'm going to open a separate issue.

pwoolvett:
> > There is a failing test for Python 2.7.
> 
> Behave is still supporting Python2 after 2 years of it's EOL date ?
> 
> I think it's time to remove support for Python2, I'm going to open a separate issue.

My thoughts exactly :rofl: 

@bittner ill take care of your suggestions later today :crossed_fingers:

pwoolvett:
can take another shot next monday

pauloxnet:
@pwoolvett thank you for your persistence in continuing to work on this PR.

I left a couple of absolutely non-blocking suggestions.

pauloxnet (approved):

bittner (approved):
Looks good to me.

I won't comment on two minor spacing issues. The effort in this PR is already significant. Congratulations! :rocket:

bittner:
Tests all pass, changes look acceptable. @jenisys I'll merge this PR now.

bittner:
@pwoolvett Thanks bunches for resisting until the end! :100: 

@pauloxnet Thank you for your help in reviewing so hard! :eyes: 

Happy holiday season, everyone! :christmas_tree:  :gift: :fireworks:

pauloxnet:
Thank you both for this merge. Now we only need a new release. 😅

pauloxnet:
We forgot to add a line in the changelog file I think.

bittner:
Please open a quick PR for that, referring to this PR. We'll have this merged directly then.

PR #1029, opened 2022-05-19
migrate to unittest.mock

### LINKED ISSUE #1028: use unittest.mock instead of mock
Could you please consider to utilize unittest.mock? It seems you support older pythons, so you probably cannot really remove mock usage, but unittest.mock could be used as a fallback.

Background:
https://trello.com/c/S6eADbii/64-remove-python-mock
https://fedoraproject.org/wiki/Changes/DeprecatePythonMock

jblanckelkoo:
need to package behave for fedora as mock is deprecated in fedora now (cf https://fedoraproject.org/wiki/Changes/DeprecatePythonMock)

Fixes #1028


jblanckelkoo:
i miss behave must still work for 2.7... :-(

PR #1039, opened 2022-06-18
Add a file with the MIT license for behave/i18n.py

musicinmybrain:
Based on:

https://github.com/cucumber/common/raw/96f5c46a95262d20418472271f556b4ba3859004/gherkin/LICENSE

### DIFF
--- a/LICENSE-i18n
+++ b/LICENSE-i18n
@@ -1,0 +1,26 @@
+The file behave/i18n.py is generated from the file gherkin-languages.json in
+the Gherkin project, under the following license:
+
+----
+
+The MIT License (MIT)
+
+Copyright (c) Cucumber Ltd, Gaspar Nagy, Björn Rasmusson, Peter Sergeant
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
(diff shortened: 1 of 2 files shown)


bittner:
Two questions:

* Is "being generated" sufficient to require including a license from the original project?
* As behave's license is already of the same type, could we simply extend the copyright notice in the project license instead of adding another file?

musicinmybrain:
> Two questions:
> 
>     * Is "being generated" sufficient to require including a license from the original project?

The usual disclaimer (I’m not a lawyer, this isn’t legal advice) applies. In the US, copyright does not protect titles, names, and short phrases, nor does it protect things like ingredient lists in recipes. However, the overall compilation of translations in the i18n file seems like it might represent sufficient creativity to qualify as a copyrightable work of authorship. If so, then

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

would apply.

It’s [helpful to distributors](https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_text), and most robust against accidental removal, if this is in the form of a license file, although a source code comment in `i18n.py` could otherwise be adequate since it would still be present in both sdists and bdists/wheels.

>     * As behave's license is already of the same type, could we simply extend the copyright notice in the project license instead of adding another file?

You could combine the two license files into a single file, but I don’t think you could combine the two notices. Besides the question of whether or not combining copyright notices is a correct approach for two projects with identical MIT licenses, behave’s license appears to be a 2-clause BSD license, with similar requirements but altogether different text.

bittner:
I believe, we should strive to become [REUSE compliant](https://reuse.software/tutorial/). REUSE introduces a LICENSES folder in a repository that maintains all licenses used in the project. In addition, all (source) files must maintain a short header with license information.

It might feel a bit cumbersome to go the REUSE route with their tooling, because it makes the code base a bit more noisy - just for adding licensing information -, but for cases like this one, for projects that have more than just one license, the pros outshine the cons, IMHO.

PR #1052, opened 2022-09-12
parser object `__init__`  not called after `reset_runtime`

mgor:
After executing `reset_runtime` and "running" behave again there would be an exception related to attributes on the `behave.matchers.ParseMatcher.parser` object.

The fix for this was to reload the `parse` module in `reset_runtime`.

Also added arguments to `reset_runtime` to selectively being able to reset specific parts of the runtime.

### DIFF
--- a/behave/runner_util.py
+++ b/behave/runner_util.py
@@ -674,6 +674,11 @@ def reset_runtime():
     matchers.ParseMatcher.custom_types = {}
     matchers.current_matcher = matchers.ParseMatcher
 
-    from importlib import reload
+    if sys.version_info.major > 2:  # reload is a builtin in 2.x
+        if sys.version_info.minor >= 4: # imp was deprecated in 3.4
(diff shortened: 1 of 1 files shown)


jenisys:
Mmh, I do not understand why a reload of the `parse` module is needed.
Could you explain me why?

NOTES:
* All `parse.Parser` objects from the `parse` module should be removed because the step registry is cleared.
* On inspection, the `parse` module has some module specific variables.
  But these should not be affected or cause any problems (as far as I can see).

mgor:
Was about to create a reproduction project to share, but right now I can't even reproduce it in the original context.

There was a bunch of `AttributeError` exceptions on `parse.Parser` when calling `load_step_modules` after calling `reset_runtime` and when troubleshooting it came down to `__init__` not being called.

But since I can't reproduce it again now, I'll close this issue.

PR #1067, opened 2022-10-30
ENHANCEMENT: Active-Tag ValueObject

jenisys:
NEW CLASS: behave.tag_matcher.ValueObject
    
* Allows the user to specific an optional comparison function,
  like: equals (default), greater_or_equal, less_or_equal, contains, ...
* Supports callable getter-function instead of current.value
* Supports type-conversion from "tag.value" as string
    
DERIVED CLASSES:
    
* NumberValueObject, BoolValueObject (in module: behave.tag_matcher)
* VersionValueObject (in module: behave.active_tag.python)
    
ACTIVE-TAG MATCHER PROTOCOL (internally used only):
    
ActiveTagMatcher class uses ValueObject class internally to support that `current_value.matches(tag_value)` can be called independent of the selected comparison function (even if no ValueObject is provided).
    
ACTIVE TAG CHANGES:
    
* ADDED: python.min_version, python.max_version
* RENAMED: python.feature.xxx (from: python_has_xxx)
    
OTHERWISE:
    
* Created "behave.active_tag" package for internal active-tag values
* Moved "behave.python" to "behave.active_tag.python"
* Moved "behave.python_feature" to "behave.active_tag.python_feature" (mostly)


jenisys:
INTEGRATED INTO: behave v1.2.7.dev2

PR #1089, opened 2023-03-07
Formatters: add steps.undefined, unused and undefinedunused

djasa:
For testing of behave repository itself, it'd be useful to have formatters that would report just faulty steps.

The StepsUsageFormatter implementing 'steps.usage' formatter already does this. This commits just adds derived classes implementing:
  * steps.undefined formatter listing just the undefined steps
  * steps.unused formatter listing just the unused steps
  * steps.unusedundefined listing both

Note:
Due to the way how behave parses steps, in case of typo in a step or param with a wrong tyep, the faulty step will appear in both listings. This is motivation for the third formatter which adds both and may be most useful for projects without many unused steps. :)

PR #1090, opened 2023-03-08
Formatters: allow steps.usage to select what to print

djasa:
For testing of behave repository itself, it'd be useful to have formatters that would report just faulty steps. The steps.usage formatter now read 'steps.usage.classes' user config key and if it's content is comma-separated list of (used|unused|undefined). If it is not recognized this way, all the step classes are printed to maintain the present behaviour for user who could already depend on this very key.

Example usage:
$ behave --dry -f steps.usage -D steps.usage.classes=unused,undefined

### DIFF
--- a/behave/formatter/steps.py
+++ b/behave/formatter/steps.py
@@ -334,9 +334,12 @@ class StepsCatalogFormatter(StepsDocFormatter):
 class StepsUsageFormatter(AbstractStepsFormatter):
     """
     Provides formatter class that shows how step definitions are used by steps.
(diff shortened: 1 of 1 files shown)


djasa:
Alternative to #1089. Example use [is there](https://github.com/behave/behave/pull/1089#issuecomment-1459070368).

The part that I don't really like is a usage of `-D`/`config.userdata`. In the parser however, I didn't see any generic infrastructure that could be used for per-formatter configs and I didn't want to hardcode the options in `--junit-directory` style.

PR #1106, opened 2023-05-04
Add instructions for extras when installing from GitHub

bittner:
How to install extras when installing from source (GitHub) is not documented yet.

Interestingly, the suggested syntax works (to make Pip not complain) and extras are installed for, e.g., `[docs]` or `[develop]`, but not for `[toml]` . Is there something wrong with [the extras declaration](https://github.com/behave/behave/blob/main/setup.py#L140-L143) maybe? Is it too complex and not resolved as expected? We may need to open a separate issue for that.

Cc @pwoolvett

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -35,33 +35,36 @@ enter the newly created directory "behave-<version>" and run::
     pip install .
 
 
-Using the Github Repository
+Using the GitHub Repository
 ---------------------------
 
 :Category: Bleeding edge
 :Precondition: :pypi:`pip` is installed
 
 Run the following command
-to install the newest version from the `Github repository`_::
-
+to install the newest version from the `GitHub repository`_::
 
     pip install git+https://github.com/behave/behave
 
-To install a tagged version from the `Github repository`_, use::
+To install a tagged version from the `GitHub repository`_, use::
 
     pip install git+https://github.com/behave/behave@<tag>
 
 where <tag> is the placeholder for an `existing tag`_.
 
-.. _`Github repository`: https://github.com/behave/behave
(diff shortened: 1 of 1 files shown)


pwoolvett:
@bittner

`tomli` is not required for python3.11, maybe you checked using that?

just cheked (devcontainer+virtualenv)

* `Python 3.7.16`+`virtualenv`+`pip==23.1` -> installs `tomli==2.0.1`
* `Python 3.11.3`+`virtualenv`+`pip==23.1` -> does not install tomli (not required)

then double checked with:

```toml
# pyproject.toml
[tool.behave]
summary = true
```
 and indeed it skips the summary, as expected for py311

bittner:
> `tomli` is not required for python3.11, maybe you checked using that?

Strange. I tried with Python 3.10 and Pipenv. At least the `Pipfile` said `3.10`.

When using `venv` the old-fashioned way it all works as you describe. Thanks! :heavy_check_mark:

jenisys:
Mmh, I am not quite sure that the merged solution is really a good solution.
Python 3.10 and newer seems to require that SSL v1.1.0 is used.

SEE ALSO:
* https://peps.python.org/pep-0644/

bittner:
> Mmh, I am not quite sure that the merged solution is really a good solution (meaning: `urllib3` pinning)

I'm confident this is just a temporary measure we need to take. RTD as a platform needs (aka "will") take care of that. Once they do we'll most likely be able to remove the version pinning. We could create an issue to get reminded later.

(In other repositories I maintain I left the docs in the current broken state, waiting for RTD to fix this issue on their side.)

PR #1113, opened 2023-05-25
Do not run before_all/after_all hooks if no features are active

dpiegdon:
Behave runs the before_all() and after_all() hooks, even if no features at all are to be executed (e.g. by filtering out all scenarios via tags).
This can be quite costly time-wise and annoying, as e.g. fixtures might be instantiated and configured in before_all/after_all.

This naïve patch fixes this by simply checking if any features are active before executing hooks.
I am open to suggestions if you think this approach has issues.


jenisys:
I am just wondering how you run into this problem.
It should be an absolute corner case meaning:
If you try to run `behave` then to run it with features/tests.

NOTE:
I assume that you used a `tag-expression` that excluded everything.

RELATED TO THE CHANGES:
* I need to look into it in more detail

POTENTIAL PROBLEMS:
* `SummaryReporter` should at least report the the statistics (how many features are skipped).
   This is currently not the case (I think) with your solution.
* Besides `select-by-tag-expression` there are other `select-by` mechanisms to select/exclude scenarios (like: `select-by-name/name-pattern`, etc.). I need to check if really everything is covered by your solution.

jenisys:
Mmh, you did not run the unit tests with `pytest` on your changes, I think ?!?

dpiegdon:
> Mmh, you did not run the unit tests with `pytest` on your changes, I think ?!?

No, sorry. I also have to admit I ported those changes from an old ubuntu VM from work, where behave v1.2.6 is installed.

> I am just wondering how you run into this problem. It should be an absolute corner case meaning: If you try to run behave then to run it with features/tests.

I suppose it is :).
I have a setup where multiple embedded devices are to be tested for multiple features-sets. Each feature-set consists of several features. To keep the gherkin readable, each feature-set has its own step-definitions related to this specific feature-set.
This is done by calling `behave` several times for each feature set. Some feature-sets may thus be completely disabled when selecting specific tags. For those I want to avoid spinning up complex fixtures like Windows programs in a virtual machine that interact with those device.

> I assume that you used a tag-expression that excluded everything.

Exactly, yes.

> POTENTIAL PROBLEMS

You are right for both cases. For `SummaryReporter` I suppose it might be easier to only guard the callback-calls with the `any(...)`, instead of discontinuing the whole function.

As for the other select-by methods, I am unsure how to proceed there, as i haven't used these. Maybe you have a better suggestion how to solve this problem?

Thanks!

PR #1118, opened 2023-06-12
Added matching keywords form gherkin-languages.json for PL

TomaszSzyborski:
When working with `#language: pl` there were missing keywords that are picked up by Gherkin plugin in PyCharm, but weren't picked up by behave stating as those keywords were unrecognized.

in version 1.2.6 behave/i18n.py when installed by pip those keywords are missing, despite being committed in 2019.

Updated:
.attic/i18n.py
basing on
etc/gherkin/gherkin-languages.json

### DIFF
--- a/.attic/i18n.yml
+++ b/.attic/i18n.yml
@@ -454,16 +454,17 @@
 "pl":
   name: Polish
   native: polski
-  feature: Właściwość
+  feature: "Właściwość|Funkcja|Aspekt|Potrzeba biznesowa"
   background: Założenia
-  scenario: Scenariusz
+  scenario: "Scenariusz|Przykład"
   scenario_outline: Szablon scenariusza
   examples: Przykłady
   given: "*|Zakładając|Mając"
-  when: "*|Jeżeli|Jeśli"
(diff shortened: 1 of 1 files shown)


jenisys:
Mmh, wrong procedure.
`behave` uses the official `cucumber gherkin-languages.json` file.
Therefore, use the following procedure.

**PROCEDURE:**
* Make pull-request on 
* After it is merged, I pull the new version of `gherkin-languages.json` and generate `behave/i18n.py` from it
* OPTIONAL: Give an info that it is merged (if I am missing this state-change)

SEE ALSO:
* [behave: etc/gherkin](https://github.com/behave/behave/tree/main/etc/gherkin)

jenisys:
NOTE:
* The `gherkin-languages.json` file was updated several time by now (since `behave v1.2.6`).
* If your changes are already contained in the `gherkin-languages.json` file,
  You should use a newer `behave version`, like `behave v1.2.7.dev3` from the git repository.

SEE ALSO:
* https://github.com/behave/behave/releases/tag/v1.2.7.dev3

PR #1132, opened 2023-08-31
Enable usage of variables in feature steps

gyoganathan:
We have a use case, where we would like to enable passing variables in the step name which will get replaced to the actual values from the variable's map we have in the before_step implementation.

Eg: 

```
When I send an email to "foo@bar.com" with subject "Hello" and body "Try out this website at http://google.com"
        And I click the link in the email I received at "foo@bar.com"
```

We want to have the ability to generate this email address dynamically for every run and use that email throughout the other steps. So we would like to do something like:

``` 
# The below step will create a random email address and store it in a context.random map as context.random["foo"] = generate_new_random_email()
Given a random email address named "foo" 
When I send an email to "{foo}" with subject "Hello" and body "Try out this website at http://google.com"
        And I click the link in the email I received at "{foo}"
```

```
def before_step(context, step):
    if 'random' in context:
        new_name = step.name.format(**context.random)
        step.name = new_name
```

This doesn't execute the step after replacement, because the match object is being built before the before_step hook is called. Moving that code before we create the match object makes it work perfectly.


jenisys:
* Placeholder translation can be easily done by using type-converters in your steps (and you can do that yourself and not need an `before_step()` preparation because your step implementation should do the work

OTHERWISE:
* we do not two things at the same time: Changing Gherkin translation and adding something new
* Gherkin keyword translations are provided in the `cucumber` repo. `behave` only pulls the latest changes and generates the source code here.

SEE ALSO:
* Gherkin i18n language details: https://github.com/behave/behave/tree/main/etc/gherkin

PR #1144, opened 2023-10-12
Update formatters.rst

DingLin23:
Adding document example for html-pretty package

### DIFF
--- a/docs/formatters.rst
+++ b/docs/formatters.rst
@@ -100,14 +100,51 @@ More Formatters
 
 The following contributed formatters are currently known:
 
-============== =========================================================================
-Name           Description
-============== =========================================================================
-allure         :pypi:`allure-behave`, an Allure formatter for behave.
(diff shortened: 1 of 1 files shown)


bittner:
Thank you for starting this PR!

I'm worried that the content you want to add is something that should really go into the installation instructions of the project itself. And in fact, [it already is](https://pypi.org/project/behave-html-pretty-formatter/).

But I noticed that [behave-html-pretty-formatter](https://pypi.org/project/behave-html-pretty-formatter/) is not in the table of the "currently known" formatters. I think, it would help to add it there. – Would you want to modify your PR in a way that it simply updates the above table instead?

bittner:
@DingLin23 Please, only update the table, hence remove the usage description of the html-pretty-formatter. Also, restore the formatting of the table, please (i.e. don't add additional spacing between the rows).

Thank you for your understanding!

PR #1151, opened 2023-12-05
Multiple, nested steps directories under steps

dron4ik86:
Updated `load_step_definitions` method to recursively search and load step definition files from nested directories within features/steps. This ensures all step files, even those deeply nested, are correctly recognized and loaded by Behave.
Previously, Behave was skipping feature files in nested structures like this:
```
features/
    ├── some_feature.feature
    └── steps/
        └── sub_folder_1/
            └── subfolder/
                └── nested_steps.py (Not executed properly)
```
Enhanced step discovery now encompasses all nested directories, ensuring availability of step definitions at any nesting level for feature tests. Corresponding tests have been added to validate this enhancement.

### DIFF
--- a/behave/runner.py
+++ b/behave/runner.py
@@ -919,12 +919,22 @@ class Runner(ModelRunner):
             self.hooks["before_all"] = self.before_all_default_hook
 
     def load_step_definitions(self, extra_step_paths=None):
+        """
+        Allow steps to import other stuff from the steps dir
+        NOTE: Default matcher can be overridden in "environment.py" hook.
+        Collect directories containing step files
+        """
         if extra_step_paths is None:
             extra_step_paths = []
-        # -- Allow steps to import other stuff from the steps dir
-        # NOTE: Default matcher can be overridden in "environment.py" hook.
+
         steps_dir = os.path.join(self.base_dir, self.config.steps_dir)
-        step_paths = [steps_dir] + list(extra_step_paths)
+        step_paths = set()
+
+        for root, _, files in os.walk(steps_dir):
+            step_paths.add(root)
+
+        step_paths = list(step_paths) + list(extra_step_paths)
+
         load_step_modules(step_paths)
 
     def feature_locations(self):
--- a/features/step.nested_folders.feature
+++ b/features/step.nested_folders.feature
@@ -1,0 +1,14 @@
+Feature: Identically named folders and modules in nested directories are all discovered
+
+Scenario: Steps are found even in identically-named folders, with differing or identical module names
+    When a step is implemented in the root folder
+    And a step is implemented in a sub folder
+    And a step is implemented in a sub-sub folder
+    And a step is implemented in another sub folder
+    And a step is implemented in another sub-sub folder with an identical name
+    And a step module with an identical name exists in the root folder
+    And a step module with an identical name exists in a sub folder
+    And a step module with an identical name exists in a sub-sub folder
+    And a step is implemented in a sub-sub-sub folder with no module in the intermediate level
+    And a step module with an identical name exists in the sub-folder two levels deeper
+    Then the implementation from all nested step modules has been executed
--- a/features/steps/nested_folders_for_steps.py
+++ b/features/steps/nested_folders_for_steps.py
@@ -1,0 +1,25 @@
+from behave import then, when
+
+
+@when('a step is implemented in the root folder')
+def step_impl(context):
+    context.nested_root_visited = True
+
+
+@when('a step module with an identical name exists in the root folder')
+def step_impl(context):
+    context.nested_step_module_in_root_visited = True
+
+
+@then('the implementation from all nested step modules has been executed')
+def step_impl(context):
+    assert context.nested_root_visited
+    assert context.nested_sub_folder1_visited
+    assert context.nested_sub_folder1_sub_folder_visited
+    assert context.nested_sub_folder2_visited
+    assert context.nested_sub_folder2_sub_folder_visited
+    assert context.nested_step_module_in_root_visited
+    assert context.nested_step_module_in_sub_folder1_visited
+    assert context.nested_step_module_in_sub_folder1_sub_folder_visited
+    assert context.nested_sub_folder3_visited
+    assert context.nested_sub_sub_folder3_visited
--- a/features/steps/sub_folder1/nested_folders_for_steps.py
+++ b/features/steps/sub_folder1/nested_folders_for_steps.py
@@ -1,0 +1,5 @@
+from behave import when
+
+@when('a step module with an identical name exists in a sub folder')
+def step_impl(context):
+    context.nested_step_module_in_sub_folder1_visited = True
--- a/features/steps/sub_folder1/sub_folder/nested_folders_for_steps.py
+++ b/features/steps/sub_folder1/sub_folder/nested_folders_for_steps.py
@@ -1,0 +1,5 @@
+from behave import when
+
+@when('a step module with an identical name exists in a sub-sub folder')
+def step_impl(context):
+    context.nested_step_module_in_sub_folder1_sub_folder_visited = True
--- a/features/steps/sub_folder1/sub_folder/sub_folder1_steps.py
+++ b/features/steps/sub_folder1/sub_folder/sub_folder1_steps.py
@@ -1,0 +1,5 @@
+from behave import when
+
+@when('a step is implemented in a sub-sub folder')
+def step_impl(context):
+    context.nested_sub_folder1_sub_folder_visited = True
--- a/features/steps/sub_folder1/sub_folder1_steps.py
+++ b/features/steps/sub_folder1/sub_folder1_steps.py
@@ -1,0 +1,5 @@
+from behave import when
+
+@when('a step is implemented in a sub folder')
+def step_impl(context):
+    context.nested_sub_folder1_visited = True
--- a/features/steps/sub_folder2/sub_folder/sub_folder2_steps.py
+++ b/features/steps/sub_folder2/sub_folder/sub_folder2_steps.py
@@ -1,0 +1,5 @@
+from behave import when
+
+@when('a step is implemented in another sub-sub folder with an identical name')
+def step_impl(context):
+    context.nested_sub_folder2_sub_folder_visited = True
--- a/features/steps/sub_folder2/sub_folder2_steps.py
+++ b/features/steps/sub_folder2/sub_folder2_steps.py
@@ -1,0 +1,5 @@
+from behave import when
+
+@when('a step is implemented in another sub folder')
+def step_impl(context):
+    context.nested_sub_folder2_visited = True
--- a/features/steps/sub_folder3/sub_folder/sub_sub_folder/sub_folder3_steps.py
+++ b/features/steps/sub_folder3/sub_folder/sub_sub_folder/sub_folder3_steps.py
@@ -1,0 +1,5 @@
+from behave import when
+
+@when('a step is implemented in a sub-sub-sub folder with no module in the intermediate level')
+def step_impl(context):
+    context.nested_sub_sub_folder3_visited = True
--- a/features/steps/sub_folder3/sub_folder3_steps.py
+++ b/features/steps/sub_folder3/sub_folder3_steps.py
@@ -1,0 +1,5 @@
+from behave import when
+
+@when('a step module with an identical name exists in the sub-folder two levels deeper')
+def step_impl(context):
+    context.nested_sub_folder3_visited = True


behave/runner.py
```diff
@@ -918,11 +918,17 @@ def load_hooks(self, filename=None):
...
+        # Load step modules from subdirectories within the steps directory
         steps_dir = os.path.join(self.base_dir, self.config.steps_dir)
-        step_paths = [steps_dir] + list(extra_step_paths)
-        load_step_modules(step_paths)
+        for root, dirs, files in os.walk(steps_dir):
+            for file in files:
+                if file.endswith('.py') and not file.startswith('__init__'):
+                    step_file = os.path.join(root, file)
+                    if os.path.isfile(step_file):  # Ensure it's a file
+                        load_step_modules([root])  # Load from the directory
```
bittner, line 931: The nested for-loop that calls [load_step_modules](https://github.com/behave/behave/blob/main/behave/runner_util.py#L566) is much too complex. It's incredibly hard to understand what it does.

Can't you collect the list of `root` folders and submit that list to `load_step_modules` in a single call?

bittner (changes requested):
What happens when you have modules of the same name in different levels of the hierarchy? Will the names clash?

@jenisys Any thoughts or comments?

behave/runner.py
dron4ik86: @bittner 
I’ve addressed the feedback by simplifying the logic. 
Now, I collect all root directories containing step files and pass them to load_step_modules in a single call.

dron4ik86:
> What happens when you have modules of the same name in different levels of the hierarchy? Will the names clash?
> 
> @jenisys Any thoughts or comments?

@bittner 
About your question about name clashes between modules with the same name in different levels of the hierarchy. 
I don’t see a realistic scenario where this would happen, given that we control the folder structure and typically avoid duplicate module names.

behave/runner.py
```diff
@@ -918,10 +918,21 @@ def load_hooks(self, filename=None):
...
         steps_dir = os.path.join(self.base_dir, self.config.steps_dir)
-        step_paths = [steps_dir] + list(extra_step_paths)
+        step_paths = set()  # Use a set to avoid duplicate paths
+
+        for root, _, files in os.walk(steps_dir):
+            if any(file.endswith('.py') and not file.startswith('__init__') for file in files):
+                step_paths.add(root)  # Add root directory containing .py step files
+
+        # Convert set to list and include extra_step_paths
+        step_paths = list(step_paths) + list(extra_step_paths)
```
bittner, line 933: I like this code much better now! The last line is basically the same as the original line 924, the for-loop is straightforward and easy to understand. Technically, the `any()` hides the nested for-loop, but at least it's a pleasure to read.

On a side-comment, I don't think we don't need all those - pretty obvious - comments.

I'm still not sure, though, why we need to run the check over all files (to exclude directories that don't have Python source code). Also, is it a good idea to exclude `__init__.py` modules? Some people might want to put actual step code into it. -- Can you explain the underlying motivation?

behave/runner.py
```diff
@@ -918,10 +918,21 @@ def load_hooks(self, filename=None):
     def load_step_definitions(self, extra_step_paths=None):
         if extra_step_paths is None:
             extra_step_paths = []
+
         # -- Allow steps to import other stuff from the steps dir
         # NOTE: Default matcher can be overridden in "environment.py" hook.
+        # Collect directories containing step files
         steps_dir = os.path.join(self.base_dir, self.config.steps_dir)
-        step_paths = [steps_dir] + list(extra_step_paths)
+        step_paths = set()  # Use a set to avoid duplicate paths
```
bittner, line 926: I'm not sure if that couldn't lead to unpleasant side-effects. Could you add a test that proves that two subdirectories having the same name but on different directory levels don't cancel out each other?

features/steps/sub_folder_1/sub_folder/nested_steps.py
```diff
@@ -0,0 +1,13 @@
...
+@given('a general step is executed')
+def step_impl(context):
+    context.general_step_executed = True
+
+
+@then('the general step should be recognized')
+def step_impl(context):
+    assert context.general_step_executed is True
+
+
```
bittner, line 13: The two trailing empty lines could be removed, I guess.

bittner (changes requested):

bittner:
> @bittner About your question about name clashes between modules with the same name in different levels of the hierarchy. I don’t see a realistic scenario where this would happen, given that we control the folder structure and typically avoid duplicate module names.

Just for comparison, Pytest has this as a [known issue](https://github.com/pytest-dev/pytest/issues/6966) with namespace packages. The solution typically is to add `__init__.py` files to folders, which makes the modules accessible via their dotted path. It would be great if you could add tests that proved the behavior we get.

Imagine, someone wanted a directory structure like this:

```
behave/
├── given
│   └── something.py
├── then
│   └── something.py
└── when
    └── something.py
```

Make sure code from all `something` modules get executed.

behave/runner.py
dron4ik86: Originally, I included this check to ensure we only collect directories with valid Python step files, avoiding unnecessary or irrelevant ones. However, after considering your feedback, I agree that the check might not be essential, so I’ve removed it.

behave/runner.py
dron4ik86: I added a test to verify that they don't cancel each other out, and the test passed without issues.

jenisys:
I am just curious why you need this suggested functionality at all.

* If you have such a high complexity with your steps, why are you not [using step-libraries] ?

[using step-libraries]: https://github.com/behave/behave/blob/main/features/step.use_step_library.feature

NOTES:

* The advantage of step-libraries is that you can verify that your step implementations work before you use them in a feature file.

behave/runner.py
jenisys: * This approach leads to problems when the same `basename` of a Python module is used twice (as mentioned by @bittner above).
* The approach that you control the module naming makes it somewhat better, but does not solve this problem. When this constraint is violated by people that are unaware of this constraint, I wish you that you can resolve the problem in a few hours or days
* THEREFORE: This is not a valid solution as long as this problem is not solved.

dron4ik86: @jenisys In my last [commit](https://github.com/behave/behave/pull/1151/commits/8d85bc15243e617ae39f38552a3818e0a864d5f8) I believe I solved the issue. 
I also created a test for the scenario where we have a directory structure like this:
```
behave/
├── given
│   └── something.py
├── then
│   └── something.py
└── when
    └── something.py
```
This is as @bittner requested in this [comment](https://github.com/behave/behave/pull/1151#issuecomment-2438950824), and the tests have passed without any issues.

bittner (changes requested):
We still need to verify that sub(sub)folders having the same name are not collapsed into a single one. You could simply rename the folders `sub_sub_folder_{1,2,3}` into just `sub_sub_folder`, and add a scenario that does the verification, like so:

```feature
    Given nested sub sub folders have the same name
    And all step modules in the nested sub sub folders have the same name
    Then the general step should be recognized in sub sub folder 1
    And the general step should be recognized in sub sub folder 2
    And the general step should be recognized in sub sub folder 3
```

That would also ensure that all 3 step modules have the same name in the sub-sub-folders. Alternatively, split the scenario up into 2 scenarios, which make it more obvious that we test 2 different aspects of the implementation.

What do you think?

bittner:
>     * If you have such a high complexity with your steps,

I'm not sure that it's really about test setup complexity. Some people might simply prefer to use subdirectories to structure the setup out of personal taste.

If we make that technically possible and there is no negative side-effect, why not? (Side note: I might be one such person that has that personal taste. :smile_cat:)

features/nested_test_2.feature
```diff
@@ -0,0 +1,5 @@
+Feature: General Feature Test - Nested Test 2
+
+  Scenario: Test if general steps are loaded from the 'nested_steps' subdirectory
+    Given a general step is executed in sub sub folder 2
```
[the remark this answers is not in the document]
dron4ik86: Thanks for the feedback! 
My main goal was just to verify that the steps are passing correctly and ensure there are no conflicts between folders, rather than strictly following the Given-When-Then structure.
That said, I can update it to "When" if it makes more sense from a BDD perspective. 
Appreciate your help!

behave/runner.py
```diff
@@ -916,12 +916,22 @@ def load_hooks(self, filename=None):
...
             extra_step_paths = []
-        # -- Allow steps to import other stuff from the steps dir
-        # NOTE: Default matcher can be overridden in "environment.py" hook.
+
         steps_dir = os.path.join(self.base_dir, self.config.steps_dir)
-        step_paths = [steps_dir] + list(extra_step_paths)
+        step_paths = set()
+
+        for root, _, files in os.walk(steps_dir):
+            step_paths.add(root)
```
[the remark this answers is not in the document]
dron4ik86: Thanks for the suggestion! I’ve already updated the code to use `os.scandir`, and everything is working as expected.

dron4ik86:
> We still need to verify that sub(sub)folders having the same name are not collapsed into a single one. You could simply rename the folders `sub_sub_folder_{1,2,3}` into just `sub_sub_folder`, and add a scenario that does the verification, like so:
> 
> ```
>     Given nested sub sub folders have the same name
>     Then the general step should be recognized in sub sub folder 1
>     And the general step should be recognized in sub sub folder 2
>     And the general step should be recognized in sub sub folder 3
> ```
> 
> We should also ensure with another scenario that all 3 step modules have the same name in the sub-sub-folders for the test we mentioned earlier, e.g.
> 
> ```
>     Given all step modules in the nested sub sub folders have the same name
>     Then the general step should be recognized in sub sub folder 1
>     And the general step should be recognized in sub sub folder 2
>     And the general step should be recognized in sub sub folder 3
> ```
> 
> What do you think?

Thanks for the suggestions! 
I’ve made the updates as you recommended.
Everything is working as expected

behave/runner.py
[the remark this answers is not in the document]
bittner: Sorry, this was meant as a suggestion for a future version, when we drop Python 2 support. For the moment we need to stick to `os.walk`, I'm afraid. The tests for Python 2.7 are failing. :disappointed:

bittner (changes requested):

features/nested_test_4.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Verify correct handling of identically named subfolders
```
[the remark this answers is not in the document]
dron4ik86: Agree with you

features/nested_test_4.feature
```diff
@@ -0,0 +1,13 @@
...
+    When a general step is executed in parent folder 1 / sub sub folder
+    Then the general step should be recognized in parent folder 1 / sub sub folder
+
+  Scenario: Test for sub_sub_folder in parent_folder_2
+    When a general step is executed in parent folder 2 / sub sub folder
+    Then the general step should be recognized in parent folder 2 / sub sub folder
+
+  Scenario: Test for sub_sub_folder in parent_folder_3
+    When a general step is executed in parent folder 3 / sub sub folder
+    Then the general step should be recognized in parent folder 3 / sub sub folder
```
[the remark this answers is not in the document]
dron4ik86: Thank you for the detailed feedback. 
I have implemented all the suggested changes. 
Each scenario has been collapsed as per your recommendation, and all steps are now correctly structured and covered across nested folders. 
I’ve also ensured that both identical and different module names are properly tested.
Please let me know if you spot anything else!

behave/runner.py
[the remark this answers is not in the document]
dron4ik86: Got it! I’ve reverted back to using `os.walk` to maintain compatibility with Python 2.7. 
The tests should pass now. 
Let me know if anything else is needed.

features/nested_test_4.feature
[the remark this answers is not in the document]
bittner: Could you go over the file names in my suggestion above again? The step module in the sub folder should likely be called `sub_folder*.py` (not `sub_sub_folder*.py`), etc. As a side note on the naming, personally I prefer `subfolder1` over `sub_folder_1`. It's more concise and easier to understand when combined in a string denoting several components (e.g. `subfolder1_subsubfolder_filename`).

Please, also rename the feature file to `step.nested_folders.feature` to match it with the feature file names that currently exist.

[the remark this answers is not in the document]
dron4ik86: I’ve made the changes as per your suggestions.
Let me know if anything else needs adjustment.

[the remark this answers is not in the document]
dron4ik86: @bittner 

In addition to the scenarios you mentioned, I’ve also added a new set of tests to cover further edge cases, particularly with sub-folder3. The goal is to ensure that steps implemented at different levels of sub-folder3 (and with identical names) are properly recognized and executed.

```
-sub_folder3/
  |-sub_folder/
    |-sub_sub_folder/
      |-sub_folder3_steps.py
-sub_folder3_steps.py
```

The goal here was to ensure that the recursive collection mechanism still works even with additional nested folders and identical names across different layers of sub-folder3. With these steps added, I believe we now have a more comprehensive set of tests that cover the edge cases for folders, subfolders, and identically named step modules.

Let me know what you think or if anything else comes to mind that we should include!

features/step.nested_folders.feature
```diff
@@ -0,0 +1,14 @@
...
+    When a step is implemented in the root folder
+    And a step is implemented in a sub folder
+    And a step is implemented in a sub-sub folder
+    And a step is implemented in another sub folder
+    And a step is implemented in another sub-sub folder with an identical name
+    And a step module with an identical name exists in the root folder
+    And a step module with an identical name exists in a sub folder
+    And a step module with an identical name exists in a sub-sub folder
+    And a step module is implemented in sub-folder3 in sub-sub folder
+    And a step module with an identical name exists in a sub-folder3
```
bittner, line 13: It's not clear to me what the intention of that is. Can you iterate on that?

(Note, I have read your explanation below, but I don't understand the alleged edge case.)

Note that the features file is all worded with generic descriptions of the problems ("identical name", "the root folder", "a subfolder"), while the step implementations have the technical details ("subfolder1", "subfolder2", etc.).

IMHO, Gherkin is for "business people" (e.g. non-technical people), while the code is for us developers.

bittner (changes requested):
I appreciate your effort! We're getting closer and closer.

@jenisys Want to take another look?

features/step.nested_folders.feature
dron4ik86: The intention behind these additional steps is to ensure that the framework provides flexibility for developers when organizing their step implementations across directories. The goal is to cover cases where identical step module names are used at different levels of nested folders—this ensures that no matter how a team structures their code, the framework can still detect and execute all steps correctly.

I understand that Gherkin is meant to be readable by non-technical stakeholders, so my focus with these particular additions was more on technical flexibility. Developers might prefer different ways of organizing their steps, whether they follow a flat structure or use deeply nested directories. The tests ensure the framework behaves consistently across both scenarios.

In essence, these additions are to give developers confidence that the recursive step collection will not break, regardless of how they choose to organize their code. If this still isn't clear or needs further alignment with the Gherkin intent, I'm happy to iterate and adjust.

Let me know what you think!

bittner: TL;DR

Don't explain. Fix the code! – Nobody puzzled about this feature scenario in future will ever come back to read this comment, likely.

If it's about recursive step collection the test should probably read, "When a step is collected recursively" or something similar.

dron4ik86: I’ve updated the scenario to reflect recursive step collection under sub-folder3, with clearer wording as per your feedback.
Let me know if it looks good!

features/step.nested_folders.feature
```diff
@@ -0,0 +1,14 @@
...
+    When a step is implemented in the root folder
+    And a step is implemented in a sub folder
+    And a step is implemented in a sub-sub folder
+    And a step is implemented in another sub folder
+    And a step is implemented in another sub-sub folder with an identical name
+    And a step module with an identical name exists in the root folder
+    And a step module with an identical name exists in a sub folder
+    And a step module with an identical name exists in a sub-sub folder
+    And a step are collected recursively from nested folders under sub-folder3
+    And a step with identical names from different levels in sub-folder3 should be executed successfully
```
bittner, line 13: I guess you intend to specify this:

```suggestion
    And a step is implemented in a sub-sub-sub folder with no module in the intermediate level
    And a step module with an identical name exists in the sub-folder two levels deeper
```

The way you have them currently worded they're more like _Then_ steps, but the step implementation shows that they are both in fact _When_ steps (there is no `assert`).

bittner (changes requested):

bittner:
When you're done with your changes, please squash all commits into a single commit - if you care about your name being attached to this PR alone. (Otherwise, when we merge the PR using the "Squash and merge" there will be the name of the person performing the squashing on the squashed commit next to yours.)

features/step.nested_folders.feature
dron4ik86: Hi, done.

bittner:
Great! I think you made it, congratulations! :1st_place_medal: 

Only one thing: The commit message now says, "Update formatters.rst" and the commit seems from someone else. Something went wrong with the squashing. Can you fix that?

features/steps/nested_folders_for_steps.py
```diff
@@ -0,0 +1,23 @@
+from behave import then, when
+
+
+@when('a step is implemented in the root folder')
+def step_impl(context):
+    context.nested_root_visited = True
+
```
bittner, line 7: The linting and test don't run anymore. There's something wrong. Please make sure checks and tests run.

```suggestion


```

Here, there should be two lines between the first, the second and the third function, instead of only one.

features/steps/nested_folders_for_steps.py
```diff
@@ -0,0 +1,23 @@
...
+@when('a step module with an identical name exists in the root folder')
+def step_impl(context):
+    context.nested_step_module_in_root_visited = True
+
```
bittner, line 11:
```suggestion


```

Same here.

bittner (changes requested):

dron4ik86:
> Great! I think you made it, congratulations! 🥇
> 
> Only one thing: The commit message now says, "Update formatters.rst" and the commit seems from someone else. Something went wrong with the squashing. Can you fix that?

Done please check.

bittner:
> > Great! I think you made it, congratulations! 🥇
> > Only one thing: The commit message now says, "Update formatters.rst" and the commit seems from someone else. Something went wrong with the squashing. Can you fix that?
> 
> Done please check.

Why is @jenisys the author of the commit, and an "andrew" the co-author? If this is not intended can you fix the commit using `git commit --amend --reset-author` (or similar) to set yourself as an author, properly?

bittner:
@jenisys Would you mind to take a look at the implementation? Is there any technical reason why we shouldn't merge the change?

dron4ik86:
> > > Great! I think you made it, congratulations! 🥇
> > > Only one thing: The commit message now says, "Update formatters.rst" and the commit seems from someone else. Something went wrong with the squashing. Can you fix that?
> > 
> > 
> > Done please check.
> 
> Why is @jenisys the author of the commit, and an "andrew" the co-author? If this is not intended can you fix the commit using `git commit --amend --reset-author` (or similar) to set yourself as an author, properly?

Done

bittner (approved):

bittner:
I'm a bit worried about the empty "Checks (0)" tab. It says, "Workflow runs completed with no jobs". I do not like.

@jenisys Do happen to know what's going on?

bittner:
Your fork is behind this repository a significant number of commits. Can you sync it and rebase your feature branch, so that you're up-to-date? Then hopefully the tests and checks will run for this PR. :crossed_fingers: 

image

Sorry, but this is needed. If tests don't run we cannot merge the PR.

dron4ik86:
> Your fork is behind this repository a significant number of commits. Can you sync it and rebase your feature branch, so that you're up-to-date? Then hopefully the tests and checks will run for this PR. 🤞
> 
> image
> 
> Sorry, but this is needed. If tests don't run we cannot merge the PR.

Fixed.
Please check.
Screenshot 2024-11-01 at 7 52 31

bittner:
> > Your fork is behind this repository a significant number of commits. Can you sync it and rebase your feature branch, so that you're up-to-date? Then hopefully the tests and checks will run for this PR. 🤞
> > Sorry, but this is needed. If tests don't run we cannot merge the PR.
> 
> Done. Please check.

Aargh! Tests and checks still don't run. That hurts! :face_with_head_bandage: 

I can't see any settings that interfere with this PR. Not sure whether this is a feature or a bug of GitHub. I suspect that has something to do with the commit user: GitHub doesn't show your GitHub username as the committer. 

Would you mind to open another PR with the same changes to see whether the checks are triggered? We can later close this one on favor of the new one if that works out. Make sure your `git config --list` settings make sense.

While you're at it, please include the word "nested" in the commit first, e.g. "Multiple, nested steps directories under steps", so we can identify the change more easily later based on the commit message. Thanks for bearing with us!

bittner:
Closing in favor of #1210 for technical reasons (supposedly GHA infrastructure issues).

PR #1162, opened 2024-02-28
Change tutorial Given to When where the Given isn't setup, …

justintaylor:
…but appears to be the key action


jenisys:
INTEGRATED INTO: behave v1.2.7.dev7

PR #1172, opened 2024-04-25
Update fixtures.rst - Clarifying uses of the yield and return values

nbmorgan:
It took me awhile to figure out that the framework was agnostic to the yielded and returned values.  I added this section to help those who follow behind.


nbmorgan:
Misunderstanding on my part

PR #1191, opened 2024-07-25
FIX #1189: Do not strip trailing whitespace from multiline text

### LINKED ISSUE #1189: Trailing whitespace in lines of multiline text statements is stripped
Commit https://github.com/behave/behave/commit/2b6845e54131c8587461c80eaa5b3ac26ca46e7f introduced whitespace normalization when parsing feature files. The motivation was to support files with Windows line endings (CRLF). The normalization is achieved by splitting lines on Unix line endings (LF) and removing trailing any whitespace in multi-line text statements.

This normalization is a bit too aggressive: it remove and trailing whitespace (not only the extra CR character).

```python
# features/steps/multiline_text.py
import re
from behave import *

trailing_whitespace_re = re.compile(r"^.*?\s+$")

@then(r"multiline text statement has trailing whitespaces on each line")
def step_impl(context):
    for line in context.text.splitlines():
        match = trailing_whitespace_re.match(line)
        assert match, f"Line {line!r} has not trailing whitespace"
```

```cucumber
# features/multiline_text.feature
Feature: normalized whitespace handling
    Scenario: Keep trailing whitespace in multiline text
        Then multiline text statement has trailing whitespaces on each line
        """
        This is a multiline string with trailing    
        whitespace at   
        the end of each   
        line   
        """
```

```bash
$ python -m behave
USING RUNNER: behave.runner:Runner
Feature: normalized whitespace handling # features/multiline_text.feature:1

  Scenario: Keep trailing whitespace in multiline text                  # features/multiline_text.feature:2
    Then multiline text statement has trailing whitespaces on each line # features/steps/multiline_text.py:6 0.000s
      """
      This is a multiline string with trailing
      whitespace at
      the end of each
      line
      """
      Assertion Failed: Line 'This is a multiline string with trailing' has not trailing whitespace



Failing scenarios:
  features/multiline_text.feature:2  Keep trailing whitespace in multiline text

0 features passed, 1 failed, 0 skipped
0 scenarios passed, 1 failed, 0 skipped
0 steps passed, 1 failed, 0 skipped, 0 undefined
Took 0m0.000s
```

The minimal fix for this problem is to restrict the stripping of trailing whitespace to CR characters:

```diff
diff --git a/behave/parser.py b/behave/parser.py
index a37614ec25ee..6c79d249f27c 100644
--- a/behave/parser.py
+++ b/behave/parser.py
@@ -668,7 +668,7 @@ class Parser(object):
 
         # -- SPECIAL CASE: Strip trailing whitespace (whitespace normalization).
         # HINT: Required for Windows line-endings, like "\r\n", etc.
-        text_line = line[self.multiline_leading:].rstrip()
+        text_line = line[self.multiline_leading:].rstrip("\r")
         self.lines.append(text_line)
 
         # -- BETTER DIAGNOSTICS: May remove non-whitespace in execute_steps()
```

But I have the feeling that Behave should rely on the built-in `str.splitlines()` to handle line endings:

```python
diff --git a/behave/parser.py b/behave/parser.py
index a37614ec25ee..d87a997c8f18 100644
--- a/behave/parser.py
+++ b/behave/parser.py
@@ -237,7 +237,7 @@ class Parser(object):
         self.reset()
         self.filename = filename
 
-        for line in text.split("\n"):
+        for line in text.splitlines():
             self.line += 1
             if not line.strip() and self.state != "multiline_text":
                 # -- SKIP EMPTY LINES, except in multiline string args.
@@ -666,9 +666,7 @@ class Parser(object):
             self.state = "steps"    # NEXT-STATE: Accept additional step(s).
             return True
 
-        # -- SPECIAL CASE: Strip trailing whitespace (whitespace normalization).
-        # HINT: Required for Windows line-endings, like "\r\n", etc.
-        text_line = line[self.multiline_leading:].rstrip()
+        text_line = line[self.multiline_leading:]
         self.lines.append(text_line)
 
         # -- BETTER DIAGNOSTICS: May remove non-whitespace in execute_steps()
@@ -748,7 +746,7 @@ class Parser(object):
         self.statement = self.rule
         self.state = "rule"
 
-        for line in text.split("\n"):
+        for line in text.splitlines():
             self.line += 1
             if not line.strip() and self.state != "multiline_text":
                 # -- SKIP EMPTY LINES, except in multiline string args.
@@ -866,7 +864,7 @@ class Parser(object):
         self.statement = model.Scenario(filename, 0, u"scenario", u"")
         self.state = "steps"
 
-        for line in text.split("\n"):
+        for line in text.splitlines():
             self.line += 1
             if not line.strip() and self.state != "multiline_text":
                 # -- SKIP EMPTY LINES, except in multiline string args.
```

Both patches fix the issue but I haven't had time to fix up the unit tests.

kahlertl:
Fix #1189 

Commit 2b6845e54131 ("FIX: Test regression on Windows") introduced whitespace normalization when parsing feature files. The motivation was to support files with Windows line endings (CRLF). The normalization is achieved by splitting lines on Unix line endings (LF) and removing trailing any whitespace in multi-line text statements.

It turns out that this normalization is a bit too aggressive: it remove any trailing whitespace -- not only the extra CR character -- from lines within the multiline text statement.

Address the issue by changing the line handling: instead of splitting on Unix LF and cleanup leftovers inside parser actions, split the lines correctly using the built-in str.splitlines() which supports a superset of universal newlines. This has the advantage that we get consistent behavior for all kinds of line separators -- including Windows, Unix, classic Mac.

### DIFF
--- a/behave/parser.py
+++ b/behave/parser.py
@@ -237,7 +237,7 @@ class Parser(object):
         self.reset()
         self.filename = filename
 
-        for line in text.split("\n"):
+        for line in text.splitlines():
             self.line += 1
             if not line.strip() and self.state != "multiline_text":
                 # -- SKIP EMPTY LINES, except in multiline string args.
@@ -666,9 +666,7 @@ class Parser(object):
             self.state = "steps"    # NEXT-STATE: Accept additional step(s).
             return True
 
-        # -- SPECIAL CASE: Strip trailing whitespace (whitespace normalization).
-        # HINT: Required for Windows line-endings, like "\r\n", etc.
-        text_line = line[self.multiline_leading:].rstrip()
+        text_line = line[self.multiline_leading:]
         self.lines.append(text_line)
 
         # -- BETTER DIAGNOSTICS: May remove non-whitespace in execute_steps()
@@ -748,7 +746,7 @@ class Parser(object):
         self.statement = self.rule
         self.state = "rule"
 
-        for line in text.split("\n"):
+        for line in text.splitlines():
             self.line += 1
             if not line.strip() and self.state != "multiline_text":
                 # -- SKIP EMPTY LINES, except in multiline string args.
@@ -866,7 +864,7 @@ class Parser(object):
         self.statement = model.Scenario(filename, 0, u"scenario", u"")
         self.state = "steps"
 
-        for line in text.split("\n"):
+        for line in text.splitlines():
             self.line += 1
             if not line.strip() and self.state != "multiline_text":
                 # -- SKIP EMPTY LINES, except in multiline string args.
--- a/tests/unit/test_parser.py
+++ b/tests/unit/test_parser.py
@@ -984,6 +984,38 @@ Feature: Stuff
         with pytest.raises(ParserError):
             parse_feature(text)
 
+    def test_parses_multiline_text_and_preserves_trailing_whitespace(self):
+        # -- ISSUE 1189: Parser removes trailing whitespace in multiline text string.
+        doc = u'''
+Feature: Multiline
+
+  Scenario: Multiline Text with Trailing Whitespace
+    Given a multiline argument with:
+      """
+      Trailing    
+      whitespace at   
+      the end of each   
+      line      
+      """
+    Then the trailing whitespace in each line ist not stripped
+'''.lstrip()
+        feature = parse_feature(doc)
+        assert feature is not None
+        assert feature.name == "Multiline"
+        assert len(feature.scenarios) == 1
+        assert feature.scenarios[0].name == "Multiline Text with Trailing Whitespace"
+        text = "\n".join([
+            "Trailing    ",
+            "whitespace at   ",
+            "the end of each   ",
+            "line      ",
+        ])
+        # pylint: disable=bad-whitespace
+        assert_compare_steps(feature.scenarios[0].steps, [
+            ('given', 'Given', 'a multiline argument with:', text, None),
+            ('then', 'Then', 'the trailing whitespace in each line ist not stripped', None, None),
+        ])
+
 
 class TestParser4AndButSteps(object):
     def test_parse_scenario_with_and_and_but(self):


jenisys:
Explain why any trailing whitespace is of importance to you !

kahlertl:
Thanks for looking into this :blush:

The syntax for multiline syntax is very similar to Pythons multiline strings. As such, it is rather unexpected that the parser tinkers with what is _inside_ the quotation marks since it a comment. Removing the leading whitespace is consistent since it indentation is used to define logical blocks (again similar to Python itself). Hence, the Behave should follow the principle of least astonishment and do not modify multiline strings besides removing the indentation level.

The real world use case is that we have Behave scenarios that compare the output of (Windows) system commands. The output these command is out of our control. Some of them – specifically `gwmi win32_pnpentity` – print trailing whitespace (because it is a table representation). Hence, the scenario file includes multiline text with trailing whitespace. This used to work with Behave 1.2.6 but broke after an upgrade to  `1.2.7.dev5` . One could argue that https://github.com/kahlertl/behave/commit/2b6845e54131c8587461c80eaa5b3ac26ca46e7f breaks backwards compatibility. We worked around the issue by adjusting our custom step implementation to strip the whitespace as well. But I would prefer to preserve the old behaviour for the afformentioned reasons:

1. Principle of least astonishment (a multiline string should be interpreted "as is")
2. Backwards compatibility

Should I extend the commit message with the motivation?

jenisys:
Mmh, sorry to say but I disagree (as already described in in the decision to reject #1189). To repeat:

* You were relying on „undefined behavior“ and not an intended feature
* Parsers strip whitespace and simplify it also here
* From your description, you were relying on a fragile binary comparison that included the trailing-whitespace (invisible)
* You can easily normalize the output of the tool before doing the comparison. This will anyway lead to a much more reliable testing strategy (IMHO).

kahlertl:
> You were relying on „undefined behavior“ and not an intended feature
> Parsers strip whitespace […]

This is about lexical analysis. Can you provide an example of another formal language that removes trailing whitespace _inside_ a string definition? All lexers that I am aware of treating the characters between string markers as a _single_ token; they are kept untouched. As such, the new behaviour of Behave is very unintuitive and confusing. Anything inside a string definition should not be modified by a parser. As such, I don't think we are relying on "undefined behaviour" but common sense amongst formal languages.

> […] and simplify it also here

What specifically is simplified by stripping whitespace inside a string definition?

> From your description, you were relying on a fragile binary comparison that included the trailing-whitespace (invisible)

Why is a precise string comparison with output of a system command brittle?

> You can easily normalize the output of the tool before doing the comparison.

The core problem is that the stripping of whitespace breaks backwards compatibility for existing scenarios that used to work fine under 1.2.6. I am not convinced that this is undefined behaviour as explained above. Behave already has the reputation of deliberately introducing breaking changes – even in patch releases. This would be yet another example of a breaking change in a minor version.

Furthermore, it forces the _convention_ of the Behave project (strip trailing whitespace) onto all its users without an opportunity to opt-out.

jenisys:
Sorry, but I would very astonished if any other user would complain over this functionality.

* [ ] Provide one simple, compelling scenario why your problem cannot be solved with the new behavior (by using the suggestions) ?

kahlertl:
For me the whole discussion is about backwards compatibility and respecting the freedom (and uses cases) of Behave users.

I think it is very unfortunate that the change in behaviour breaks an existing use case for a Behave customer. We should do better with respect to backwards compatibility. Currently, a _patch_ version upgrade from 1.2.6 to 1.2.7 requires downstream work. This is very unexpected and reinforced the reputation of Behave deliberately breaking stuff in releases.

PR #1244, opened 2025-04-24
fix - model.py table - make line optional

CCedricYoung:
This is a very minor issue, introduced in e468869932b8edd889c5eb9c25ec77e25ea2f7bc, since the optional param has default `None` and when unset it causes the exception `TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`.

I doubt others will run into this, but posting this PR just as courtesy.

### DIFF
--- a/behave/model.py
+++ b/behave/model.py
@@ -2023,7 +2023,7 @@ class Table(Replayable):
 
         if rows:
             for index, row in enumerate(rows):
-                self.add_row(row, line+index+1)
+                self.add_row(row, self.line+index+1)
 
     def clear(self, headings=None, keep_headings=True):
         if keep_headings and not headings:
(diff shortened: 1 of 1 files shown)


behave/model.py
```diff
@@ -2023,7 +2023,7 @@ def __init__(self, headings, rows=None, line=None):
 
         if rows:
             for index, row in enumerate(rows):
-                self.add_row(row, line+index+1)
+                self.add_row(row, self.line+index+1)
```
bittner, line 2026:
```suggestion
                self.add_row(row, self.line + index + 1)
```

bittner:
Looks good, thank you! - Can you only fix the formatting as suggested?

bittner (approved):

bittner:
Thank you! :1st_place_medal:

CCedricYoung:
wow, thanks for the quick review!

PR #1249, opened 2025-06-24
Fixed having one event loop per async step for python >= 3.10

jeteve:
Hi,

This addresses an issue with using behave with async steps on python >= 3.10.

Essentially, post 3.10 , behave create a new event loop for each async steps, preventing long running test application level async loops to perform their duty correctly across multiple async steps.

The new test demonstrates the failure and will not pass without the new code.

I've made a new branch from the `1.2.7.dev-5`  tag as we use this and would like a patch to this specific branch (`1.2.7.dev-5.1` ?), feel free to port to your main branch too!

### DIFF
--- a/behave/api/async_step.py
+++ b/behave/api/async_step.py
@@ -60,7 +60,7 @@ _PYTHON_VERSION = sys.version_info[:2]
 # A global asyncio.Runner instance to use to get the single
 # event loop in the async_run_until_complete decorator.
 # Since 3.10, this is the preferred way to get the equivalent of the global event loop.
-ASYNC_IO_RUNNER = ASYNC_IO_RUNNER = asyncio.Runner() if has_asyncio and _PYTHON_VERSION >= (3, 10) else None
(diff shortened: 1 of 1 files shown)


jenisys:
Thank you for your effort:

* I have manually added your tests to the `behave` test suite.
* BUT: I will not merge this PR (and will close it).

REASON:

* I added direct support for async-step-functions to `behave` (while you were working on your solution to the problem).
* The decorator will still work but is deprecated and will be removed in a future version.

SEE ALSO:

* [features/step.async_steps.feature](https://github.com/behave/behave/blob/main/features/step.async_steps.feature)
* [features/step.async_steps.step_decorator.feature](https://github.com/behave/behave/blob/main/features/step.async_steps.step_decorator.feature)

NOTES:
* Please verify that the current solution works for you by using the current HEAD of this repository.
* I will publish a pre-release `v1.2.7.dev7` on https://pypi.org in the next few days.

Issue #1234, opened 2025-02-18
userdata_defines configuration file parameter is not up to date


vvavrychuk:
Here https://behave.readthedocs.io/en/latest/behave/#configuration-parameters documents userdata_defines parameter, but it seems to be ignored. Instead, to pass userdata via configuration file `[behave.userdata]` section should be used with userdata as `name = value` properties under this section.

nirmalnirav:
Yes this problem is still active. I had to debug the application ` python3 -m pdb -m behave`  to reach the same conclusion! 😅

bittner:
IIUC, the open `userdata_defines` is a command-line only option. That's what the source code says.

- [Setup of command line option](https://github.com/behave/behave/blob/main/behave/configuration.py#L133-L137) `[-D | --define]`
- [Declaration of list of excluded options](https://github.com/behave/behave/blob/main/behave/configuration.py#L423-L431) (used in [configfile_options_iter](https://github.com/behave/behave/blob/main/behave/configuration.py#L485), which is used in [read_configparser](https://github.com/behave/behave/blob/main/behave/configuration.py#L525-L532) and [read_toml_config](https://github.com/behave/behave/blob/main/behave/configuration.py#L573-L595))
- [Declaration of command line only set](https://github.com/behave/behave/blob/main/behave/configuration.py#L769) (runner configuration)
- [Presumably dead code for setup and update userdata](https://github.com/behave/behave/blob/main/behave/configuration.py#L1186-L1200) (only used in tests, FWICS)

Does that need to be stated explicitly in the docs, maybe?

@jenisys Can you comment on why the implementation makes sense, or if that is actually a bug?

jenisys:
As described by @bittner .
* Command line option can override data from userdata section in the config-file.

SEE: [features/userdata.feature](https://github.com/behave/behave/blob/main/features/userdata.feature)

PR #1256, opened 2025-08-08
docs: userdata

Therdel:
- address issue #1234 by updating configuration file docs
    - The other undocumented sections [behave.formatters and behave.runners](https://github.com/behave/behave/blob/v1.3.0/behave/configuration.py#L576-L577) remain
- give `Userdata` its own documentation page

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -452,6 +452,14 @@ RAW_VALUE_OPTIONS = frozenset([
     # -- MAYBE: "scenario_outline_annotation_schema",
 ])
 
+# -- SPECIAL CONFIGURATION SECTIONS
+# Maps config sections to their data names in the final config
+SPECIAL_CONFIG_SECTIONS = {
+    "formatters": "more_formatters",
+    "runners":    "more_runners", 
+    "userdata":   "userdata",
+}
+
 
 def _values_to_str(data):
     return json.loads(json.dumps(data),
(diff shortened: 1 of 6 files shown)


jenisys:
Thanks, a clear improvement.

BUT: Are you aware that `behave.rst` is generated by [update_behave_rst.py](https://github.com/behave/behave/blob/main/docs/update_behave_rst.py).
There your change will be overwritten/removed on the next run of this update-script.

Therdel:
Curious. I did `invoke docs` as written in [contributing.rst](https://github.com/behave/behave/blob/v1.3.0/docs/contributing.rst) in my PR, and my changes persisted 🤔
...
Ah. [docs.py calls sphinx-build](./tasks/docs.py)](https://github.com/behave/behave/blob/v1.3.0/tasks/docs.py#L68) directly, without calling the update script like the [Makefile for Sphinx does](https://github.com/behave/behave/blob/v1.3.0/docs/Makefile#L6).
Is this duplicated docs build code?

I merged some duplicated definitions of excluded config file options and special config sections and moved the Additional configurations section to `behave.rst-template`

jenisys:
Two points:
* I have doubts that your changes in `behave.configuration` related to `SPECIAL_CONFIG_SECTIONS` will work for INI files (but I have not checked it yet)
* English: Use uppercase Letters for substantives in section titles

Therdel:
putting the PR to Draft.

- fixed the title.
- Also, I think the Userdata doc isn't visible enough in `Noteworthy in Version 1.2.5`, so I refactored it as a main documentation page. Also, I rephrased it and adapted the structure.

jenisys:
The failures seems to be caused by the changed if-condition in `Configuration` class constructor: 

```python
# -- APPROX: Line 830ff
if key.startswith("_") or key in CONFIGFILE_EXCLUDED_OPTIONS
# -- MUST BE: Corresponding to logic-expression before
if key.startswith("_") and key not in CONFIGFILE_EXCLUDED_OPTIONS:
```

jenisys:
@Therdel 
Thanks for the effort.

PR #1264, opened 2025-08-22
fix: Configuration argument `verbose` unused

satelliteprogrammer:
Configuration is not applying the verbose it receives as argument.

The issue is that the `init` method receives `verbose` as a separate
argument from all others and is not passing it to `make_default`. The
defaults are then passed to the `ArgumentParser`, which would be missing
the verbose initially passed to `Configuration`, and where verbose has
action="store_true", making it default to False.

This patch also stores `verbose` in the defaults used by the arg parser.

### DIFF
--- a/behave/configuration.py
+++ b/behave/configuration.py
@@ -835,7 +835,7 @@ class Configuration(object):
         """
         (Re-)Init this configuration object.
         """
-        self.defaults = self.make_defaults(**kwargs)
+        self.defaults = self.make_defaults(verbose=verbose, **kwargs)
         self.version = None
         self.capture = None
         self.capture_stdout = None


jenisys:
I will think about it.
But note that the following parameter usage performs what you desire:

```python
config = Configuration([“—verbose“])
assert run_behave(config) == os.EX_OK
```

satelliteprogrammer:
And that's probably what we will use for now.
However, given that we're calling `Configuration` directly and the library offers a possibility to pass arguments directly, it's much nice to organize all options in a dictionary and then expand it when calling `Configuration`.
```python
options = {
    "junit": True,
    "load_config": False,
    # ...
    "verbose": True,
}

config = Configuration([], tags=tags, paths=[tests_path()], stage="", **options)
```

Issue #1292, opened 2025-12-16
Type checkers produce errors on step decorators


lypwig:
Both Pyright and ty complain produce a "call check" error on given/when/then decorators:
- [ty: call-non-callable](https://docs.astral.sh/ty/reference/rules/#call-non-callable)
- [pyright: reportCallIssue](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportCallIssue)

Example code:

```py
from behave import given

@given("something")  # <-- type check error
def do_something(context):
    pass
```

I had to disable type checking on all lines containing decorators (using `# type: ignore`), which is a bit tedious.

vijaymodha:
I'd like to propose the following workaround while we wait for type checking to be explicitly supported.

Consider making the `_StepDecorator` stub return a callable stub. This change to behave/behave@ee58244 (branch release/v1.3.x) allows `pyright` and `ty` to succeed:

```diff
diff --git a/behave/step_registry.py b/behave/step_registry.py
index bbe92e2..c0be348 100644
--- a/behave/step_registry.py
+++ b/behave/step_registry.py
@@ -214,7 +214,11 @@ class StepRegistry(object):
 # -----------------------------------------------------------------------------
 # -- STEP-DECORATOR PLACEHOLDERS (setup at end of this module):
 class _StepDecorator:
-    pass
+    def __call__(self, *args, **kwargs):
+       def decorator(func):
+           return func(*args, **kwargs)
+
+       return decorator


 registry = StepRegistry()
```

I use VSCode + Python Extension Pack (Pylance) which uses `pyright` under the hood.

I can reproduce my issue by calling `pyright` (and @lypwig's by calling `ty`) in a new directory (say `behave-proposal`):

1. Create a virtualenv (I'm using python 3.14) and install `behave` and `ty`.
2. Run `npm install --save pyright`.
3. Create `test.py` from [@lypwig's example code](https://github.com/behave/behave/issues/1292#issue-3733853661)
4. Run `ty check` and `pyright`:

    ```shell
    (venv) /tmp/behave-proposal $ ty check
    error[call-non-callable]: Object of type `_StepDecorator` is not callable
     --> test.py:3:2
      |
    1 | from behave import given
    2 |
    3 | @given("something")  # <-- type check error
      |  ^^^^^^^^^^^^^^^^^^
    4 | def do_something(context):
    5 |     pass
      |
    info: rule `call-non-callable` is enabled by default
    
    Found 1 diagnostic

[...]

PR #1295, opened 2025-12-28
fix: make _StepDecorator a callable stub

vijaymodha:
Implements workaround in [#1292](https://github.com/behave/behave/issues/1292#issuecomment-3694973210) before explicitly supporting type checking.

Type checkers like `pyright` and `ty` use static analysis to determine types (instead of dynamic/run-time analysis). Since the behave decorators are declared with the [`_StepDecorator` stub (placeholder)](https://github.com/behave/behave/blob/d659aab9c6510f021ab18849c6fb4f2ed15ea6a4/behave/step_registry.py#L219-L228) and then [replaced at runtime](https://github.com/behave/behave/blob/d659aab9c6510f021ab18849c6fb4f2ed15ea6a4/behave/step_registry.py#L242), both `pyright` and `ty` report that the decorator is not callable.

```shell
(venv) /tmp/behave-proposal $ ty check
error[call-non-callable]: Object of type `_StepDecorator` is not callable
 --> test.py:3:2
  |
1 | from behave import given
2 |
3 | @given("something")  # <-- type check error
  |  ^^^^^^^^^^^^^^^^^^
4 | def do_something(context):
5 |     pass
  |
info: rule `call-non-callable` is enabled by default

Found 1 diagnostic

(venv) /tmp/behave-proposal $ npx pyright
/tmp/behave-proposal/test.py
  /tmp/behave-proposal/test.py:3:2 - error: Object of type "_StepDecorator" is not callable
    Attribute "__call__" is unknown (reportCallIssue)
1 error, 0 warnings, 0 informations
```

We can correct this by changing the stub to be a callable stub. The same commands with this branch are successful:

```shell
(venv) /tmp/behave-proposal $ ty check
All checks passed!

(venv) /tmp/behave-proposal $ npx pyright
0 errors, 0 warnings, 0 informations
```

### DIFF
--- a/behave/step_registry.py
+++ b/behave/step_registry.py
@@ -212,10 +212,11 @@ class StepRegistry(object):
 # -- STEP-DECORATOR PLACEHOLDERS (setup at end of this module):
 class _StepDecorator:
     def __call__(self, *args, **kwargs):
-       def decorator(func):
-           return func(*args, **kwargs)
+        def decorator(func):
+            # Placeholder: no-op decorator, just return the original function.
+            return func
 
-       return decorator
+        return decorator
 
 
 registry = StepRegistry()


bittner:
Thank you! :1st_place_medal:

PR #1297, opened 2026-01-12
Fixed incorrect config reference in Context causing cleanup_error's

wiebren:
Fixed incorrect `config` reference in `Context` causing cleanup_error's

### DIFF
--- a/tests/unit/test_runner_context.py
+++ b/tests/unit/test_runner_context.py
@@ -98,13 +98,17 @@ class TestContext:
         """
         context = self.make_context()
 
+        # -- SIMULATE: User masks root attributes, like in a "before_all" hook.
         with context.use_with_user_mode():
-            for key in context._root.keys():
-                if key[0] != '@':
-                    context.__setattr__(key, False)
+            for key in list(context._root.keys()):
+                if key[0] != "@":
+                    setattr(context, key, False)
 
-        # -- SHOULD-NOT-RAISE:
-        context._pop()
+        # -- SHOULD-NOT-RAISE: Cleanups must use the real config object.
+        with scoped_context_layer(context, "scenario"):
+            pass
+
+        assert len(context._stack) == 1
 
 
 class TestContext2(unittest.TestCase):
--- a/tests/unit/test_runner_context.py
+++ b/tests/unit/test_runner_context.py
@@ -103,6 +103,7 @@ class TestContext:
                 if key[0] != '@':
                     context.__setattr__(key, False)
 
+        # -- SHOULD-NOT-RAISE:
         context._pop()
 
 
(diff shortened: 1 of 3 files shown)


bittner:
@wiebren I had Copilot write a test for us (#1306). Can you please take that code, clean it up a bit, and integrate it into your PR? If you don't like the code at all, fine too. You can take it as inspiration at least.

bittner:
Please rebase your branch and integrate the suggested test. I can then merge your changes.

wiebren:
@bittner sorry for the delay, I added a test case that overrides all context root attributes, that also covers this scenario. 

The entire `TestContext` class was not marked as a unit test, so I also fixed that.

tests/unit/test_runner_context.py
```diff
@@ -88,6 +88,16 @@ def test_context_contains(self):
         context._push()
         assert "thing" in context
 
+    def test_context_root_override_should_not_break_context(self):
```
bittner, line 91: Please add a concise docstring explaining the motivation for this test.

Cause + effect format would be nice (e.g. "When ... then ...").

bittner (changes requested):
Please rebase your feature branch onto our current default branch, and resolve the conflicts. Thank you!

@jenisys Any comments before this gets merged?

tests/unit/test_runner_context.py
wiebren: Added, in the cause + effect shape you asked for:

```python
"""
When a user overrides a root attribute of the context, like "config",
then popping the context layer still works.

The cleanup path has to reach the real config object, not whatever the
user happens to have stored under that name.
"""
```

wiebren:
Rebased onto current `main` and added the docstring — now at f1ec21e, and the PR is mergeable again.

Sorry for the long silence on this one.

The branch was 26 commits behind, and the rebase was clean: no conflicts, so the contribution is still the same two lines it always was.

```diff
-            if self.config.should_capture_hooks():
+            if self._config.should_capture_hooks():
```

Still relevant on current `main`, where `Context._pop` reaches the config through `self.config` at `behave/runner.py:330`. That goes through `Context`'s user attribute lookup, so any user who stores something under the name `config` shadows the real object and cleanup raises. `ModelRunner.config` (line 808) is a genuine attribute and is deliberately left alone.

The test earns its place — with the fix reverted it fails with exactly the reported symptom:

```
E   AttributeError: 'bool' object has no attribute 'should_capture_hooks'
behave/runner.py:330: AttributeError
```

`pytest tests/unit` on the rebased branch: 1408 passed, 2 skipped.

@jenisys — you were asked for a second opinion back in February, in case you still want to weigh in.

bittner:
Thanks for updating the PR! :100: 

No idea why GHA don't run. Tests pass locally, hence merging. :rocket:

PR #1301, opened 2026-01-18
remove Python2+3 compatibility shim

a-detiste:
Hi,

Here is an other cleanup for your consideration.

Greeetings

Alexandre

### DIFF
--- a/behave/__main__.py
+++ b/behave/__main__.py
@@ -5,10 +5,15 @@ import locale
 import sys
 from behave.version import VERSION as BEHAVE_VERSION
 from behave.configuration import Configuration
-from behave.exception import (ConstraintError, ConfigError,
-    InvalidFileLocationError, InvalidFilenameError,
-    ClassNotFoundError, InvalidClassError,
-    TagExpressionError)
+from behave.exception import (
+    ClassNotFoundError,
+    ConfigError,
+    ConstraintError,
+    InvalidClassError,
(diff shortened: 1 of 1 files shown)


behave/exception.py
```diff
@@ -9,7 +9,7 @@
 # -- RE-EXPORT: Exception class(es) here (provided in other places).
 #   USE MODERN EXCEPTION CLASSES: FileNotFoundError, ModuleNotFoundError
 #   COMPATIBILITY: Emulated if not supported yet by Python version.
-from behave.compat.exceptions import (FileNotFoundError, ModuleNotFoundError)  # noqa: F401
+from builtins import (FileNotFoundError, ModuleNotFoundError)  # noqa: F401
```
bittner:
```suggestion
from builtins import FileNotFoundError, ModuleNotFoundError
```

AFAIK, you don't need parentheses in this case, and Ruff doesn't complain about unused imports when the imported symbols are listed in `__all__` (even though they're just strings).

bittner (changes requested):
We likely don't need to import the built-in exception classes.

a-detiste:
I plan to remove six too in a next PR

bittner (changes requested):
Looks quite good to me. Just a small change, still.

The linter should really report code style issue. It worries me that we're not using it consistently in the project, not on GHA. That would take most of the emotion out of those issues. 😟

PR #1302, opened 2026-01-18
trim Py2-compatible class syntax

a-detiste:
In Python3 all classes are "New-Style Classes"

https://www.geeksforgeeks.org/python/comparing-old-style-and-new-style-classes-in-python/

This change was edited with this script:

```
grep '(object):$' -rl | grep -v issue.feat | while read f; do sed -i 's/(object):$/:/' $f ; done
grep '(object): pass' -rl | grep -v issue.feat | while read f; do sed -i 's/(object): pass/: pass/' $f ; done
```


bittner:
Thanks @a-detiste, great job! :100:

PR #1307, opened 2026-01-20
trim hybridation against Python < 3.5

a-detiste:
the spun-out `_*py35.py` files are merged back were they belong


behave/log_config.py
bittner: We can remove this, too. See comment below.

bittner:
```suggestion
if errors:
```

Python > 3.9 is guaranteed. According to [pyproject.toml](https://github.com/behave/behave/blob/main/pyproject.toml#L63) we only support Python 3.10+.

(GitHub's suggestion feature seems to be broken)

bin/json.format.py
```diff
@@ -44,7 +43,7 @@ def json_format(filename, indent=DEFAULT_INDENT_SIZE, **kwargs):
     :raises:  IOError (Error 2), if file not found.
     """
     console  = kwargs.get("console", logging.getLogger("console"))
-    encoding = kwargs.get("encoding", None)
+#    encoding = kwargs.get("encoding", None)
```
bittner: Why is this commented out? Shouldn't we delete it if it's not used anymore?

bin/jsonschema_validate.py
```diff
@@ -30,17 +30,13 @@
...
 
 # -----------------------------------------------------------------------------
 # FUNCTIONS:
 # -----------------------------------------------------------------------------
 def json_loads(text, encoding=None):
     kwargs = {}
-    if encoding and PYTHON_VERSION < (3, 1):
-        # -- NOTE: encoding keyword is deprecated since python 3.1
-        kwargs["encoding"] = encoding
     return json.loads(text, **kwargs)
```
bittner: I think, the whole function is now just `json.loads` and should be removed entirely. The `encoding` argument is not used anymore and `kwargs` is always empty.

tests/unit/test_deprecated.py
```diff
@@ -4,6 +4,10 @@
 * Runs tests related to async-steps.
 """
 
+from behave.api.async_step import (
+    async_run_until_complete,
+    use_or_create_event_loop,
+)
 from behave.capture import CaptureController
 from behave.configuration import Configuration
 from behave.python_feature import PythonFeature
```
bittner: I think this is an unused import now.

bittner (changes requested):
A few things I noted...

bin/jsonschema_validate.py
```diff
@@ -35,16 +35,12 @@
 # -----------------------------------------------------------------------------
 # FUNCTIONS:
 # -----------------------------------------------------------------------------
-def json_loads(text, encoding=None):
-    kwargs = {}
-    return json.loads(text, **kwargs)
-
 
 def json_load(filename, encoding=None):
```
bittner, line 39: The encoding argument is unused. Can we remove it entirely? - Options:

- Either check and adjust all uses of the `json_load` function
- Or add code that emits a (deprecation) warning when encoding is not None

tests/unit/test_warnings.py
```diff
@@ -1,5 +1,36 @@
...
+
+
+# -----------------------------------------------------------------------------
+# TEST SUITE
+# -----------------------------------------------------------------------------
+# -- REQUIRES:
+# * Python >= 3.5  -- Using async-function syntax
+# * Python <  3.11 -- NOT-SUPPORTED YET: asyncio.timeout()
+@pytest.mark.skipif(PythonLibraryFeature.has_asyncio_timeout(),
+                    reason="asyncio.timeout() is supported.")
```
bittner: Modern formatting would look like this:

```suggestion
@pytest.mark.skipif(
    PythonLibraryFeature.has_asyncio_timeout(),
    reason="asyncio.timeout() is supported.",
)
```

tests/unit/test_warnings.py
```diff
@@ -1,5 +1,36 @@
...
+class TestAsyncStepFunction:
+    def test_warning__ctor_with_timeout(self):
+        """Ensure that the ``AsyncStepFunction(..., timeout=...)`` issues a warning."""
+        with pytest.warns(RuntimeWarning, match=r"IGNORED: timeout"):
+            _ = AsyncStepFunction(some_async_func, timeout=42)
```
bittner: What does this code do? Do we need the assignment or would the function call be enough?

bittner (changes requested):

bin/jsonschema_validate.py
a-detiste: less is more

tests/unit/test_warnings.py
a-detiste: ... I have no idea

bittner: I intended: It's probably safe to remove the assignment. - It's a test and simply constructing the object will do, won't it?

```python
        with pytest.warns(RuntimeWarning, match=r"IGNORED: timeout"):
            AsyncStepFunction(some_async_func, timeout=42)
```

a-detiste: That reminds me I had the "chance" to follow a three days slop-infused course at job last year ... not learned much, especially about everything Async ...

a-detiste:
This one should be done; next one will be about `six` before I loose track

a-detiste:
@bittner  is it ok ?

behave4cmd0/log_steps.py
```diff
@@ -72,8 +71,6 @@
 )
 
 
-_PYTHON_VERSION = sys.version_info[:2]
-
```
bittner: This is an additional empty line. The formatter might complain about 3 empty lines here.

tests/unit/test_deprecated.py
```diff
@@ -4,26 +4,39 @@
 * Runs tests related to async-steps.
 """
 
+from behave.api.async_step import (
+    async_run_until_complete,
+    use_or_create_event_loop,
+)
```
bittner: The import formatted as a single line fits in the 88 character range.

```suggestion
from behave.api.async_step import async_run_until_complete, use_or_create_event_loop
```

bittner (changes requested):
Just a few nit-picks. - Sorry for having you wait for so long!

tests/unit/test_deprecated.py
a-detiste: ok

bittner:
Hmmm, I forgot to squash the commits. I should work on PRs when I'm tired. :worried: 

Thanks for your contribution, @a-detiste!

a-detiste:
I'm tired and I bought a train ticket for the wrong day 😵‍💫

"six" will wait, I m gone

PR #1309, opened 2026-02-21
finish removal of "six"

a-detiste:
I vendored 1 tiny function

### DIFF
--- a/behave/runner.py
+++ b/behave/runner.py
@@ -9,8 +9,6 @@ import traceback
 import warnings
 import weakref
 
-import six
-
 from behave._types import (
     require_type,
     require_not_type,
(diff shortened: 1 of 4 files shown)


behave/runner.py
```diff
@@ -36,6 +34,18 @@
 from enum import Enum
 
 
+def reraise(tp, value, tb=None):
```
bittner, line 37: Can you add a docstring explaining the motivation for this function?

It's not clear why "re-raise" wouldn't simply be a simple `raise` (with not arguments).

bittner (changes requested):

behave/runner.py
a-detiste: This is the snippet taken verbatim from six.py

Yes it could be simpler ... I m not sure if raise is enough.

That's why I does this PR last

bittner: I understand you've taken it from six, but we should explain its existence.

c92s: @bittner should it still be Python 2.7 compatible? If @a-detiste does not want to continue with their change, I can also create a quick PR with the removal.

a-detiste: please help, you can take over

c92s: @bittner created a quite similar PR regarding the removal [here](https://github.com/behave/behave/pull/1317).

bittner: I merged the related PR. I guess we can close this PR now?

bittner:
Thanks for your input to get this fixed via #1317. :1st_place_medal:

PR #1321, opened 2026-05-06
auto-promote raw bool values to BoolValueObject in ActiveTagMatcher

### LINKED ISSUE #1320: Feature Request: Better Boolean handling for Active Tags
Hello to the Behave team and thank you very much for your hard work! I love Behave and am using it very successfully in my workplace to automate tests on 3D printer's firmware and software!

We use the Active Tag system (documented [here](https://behave.readthedocs.io/en/latest/new_and_noteworthy_v1.2.5/#active-tags)) and I noticed something unintuitive that I think should be easy to fix:

If you set an active tag value to a normal Boolean (`True` or `False`), it doesn't seem possible to evaluate against that tag successfully. For example:

```python
# in environment.py
active_tag_value_provider = {
    "my_tag": "undefined",
}


def before_all(context):
    setup_active_tag_values(active_tag_value_provider, {"my_tag": True}
```

```feature
# in my_feature.feature
Feature: My Feature
    @use.with_my_tag=true
    Scenario: My Scenario
        Given nothing is happening
        When I do something
        Then it happens
```

In this example, `My Scenario` will never be run, because the active tag code will evaluate `my_tag  = True`, and `True != "true"`. This is the case even if you use  `@use.with_my_tag=True`

Currently I'm mitigating this by wrapping my Booleans in `BoolValueObject`, which is working, but it's not ergonomic. 

## Suggestion

It would be nice if Booleans would get automatically wrapped so that `{"my_tag": True}` and `@use.with_my_tag=true` "Just Work" together.

## Full reproduction code
I made a full code snippet to demonstrate the issue: https://github.com/gabeseltzer/active-tag-min-reproducable-code

gabeseltzer:
- fixes https://github.com/behave/behave/issues/1320

See the issue above for motivation

This PR auto-promotes raw bool provider values to BoolValueObject at the wrap site, so `@use.with_<cat>=true` and `@use.with_<cat>=false` work as users expect when the provider value is a Python bool.

Also adds tests!

(Let me know if there's additional testing I can do to make sure this works well!)

### DIFF
--- a/behave/tag_matcher.py
+++ b/behave/tag_matcher.py
@@ -51,6 +51,22 @@ class ValueObject:
         self._value = value
         self.compare = compare
 
+    @classmethod
+    def from_raw(cls, value):
+        """Wrap a raw provider value in the most appropriate ValueObject subclass.
+
+        Picks :class:`BoolValueObject` for ``bool`` values so that string tag
+        values like ``"true"`` / ``"false"`` are coerced before comparison —
+        otherwise ``operator.eq(True, "true")`` is always ``False`` and bool
+        active-tag matching silently never matches. See issue #1320.
+
+        Note: ``isinstance(value, bool)`` must be checked before any ``int``
+        check since ``bool`` is a subclass of ``int`` in Python.
+        """
+        if isinstance(value, bool):
+            return BoolValueObject(value)
+        return cls(value)
+
     @property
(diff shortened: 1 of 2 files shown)


bittner:
Thanks a lot for your contribution! :1st_place_medal:

PR #1342, opened 2026-08-14
fix: scenario_autoretry does not reset runner.hook_failures (#1128)

### LINKED ISSUE #1128: scenario_autoretry does not reset runner.hook_failures
The scenario_autoretry monkey patch does not intervene with or reset the increment of `self.hook_failures` in the runner. Consequently, if before_scenario/after_scenario fails the first time, and then passes on the retry, behave still exits with failure at the end because `hook_failures` is non-zero.

https://github.com/behave/behave/blob/1c6197b35c15e07b5bae62b3131a98f9caa88f4e/behave/runner.py#L649
https://github.com/behave/behave/blob/1c6197b35c15e07b5bae62b3131a98f9caa88f4e/behave/runner.py#L751-L755

nikolauspschuetz:
## Summary

Fixes #1128.

`patch_scenario_with_autoretry()` retried a failed scenario but never reset the
test-runner's hook-failure state. When a `before_scenario` / `after_scenario`
hook failed on one attempt and the scenario then passed on a retry, the scenario
was correctly reported as **passed**, yet `behave` still exited with **failure**
because `runner.hook_failures` was left non-zero (the exit status is computed as
`... or self.hook_failures > 0` in `ModelRunner.run_model()`).

## Fix

Follows the solution sketch from @jenisys in the issue: record the test-runner
state (only `runner.hook_failures`) before the retry loop and restore it before
each attempt, so a hook failure from a previous attempt does not leak into a
later, passing one.

```python
runner = args[0]
hook_failures_on_start = runner.hook_failures
for attempt in range(1, max_attempts + 1):
    runner.hook_failures = hook_failures_on_start   # -- restore before each (retried) run
    ...
```

## Regression test

`issue.features/issue1128.feature` — a `before_scenario` hook fails on the first
attempt and passes on the retry. It asserts the run **passes** with
`1 scenario passed, 0 failed`.

- On `main` (before the fix): the inner run reports `1 scenario passed, 0 failed`
  but exits `1` (`Expected: <0> but: was <1>`) → the feature **fails**.
- With the fix: the run exits `0` → the feature **passes**.

## Verification

Run in a clean `python:3.12` container (`pip install -e ".[testing]"`):

- `bin/behave issue.features/issue1128.feature` — RED before, GREEN after.
- `bin/behave features/runner.scenario_autoretry.feature` — still passes (no regression).
- `pytest tests/` — `1639 passed, 3 skipped, 2 xfailed`.

`CHANGES.rst` updated.

---
_fwiw: drafted with AI assistance (Claude); reviewed and verified by me._

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,7 @@ FIXED:
 * Include changes from ``behave v1.3.1`` (#1255, #1239)
 * issue #1028: use unittest.mock instead of mock (submitted by: pgajdos)
 * pull #1297: Context._pop(): Use internal config reference to avoid cleanup-errors if "context.config" is masked (provided by: wiebren)
+* issue #1128: scenario_autoretry now resets ``runner.hook_failures`` between retry attempts
 
 DOCUMENTATION:
 
--- a/behave/contrib/scenario_autoretry.py
+++ b/behave/contrib/scenario_autoretry.py
@@ -48,7 +48,16 @@ def patch_scenario_with_autoretry(scenario, max_attempts=3):
     :param max_attempts:    How many times the scenario can be run.
     """
     def scenario_run_with_retries(scenario_run, *args, **kwargs):
+        # -- HINT: Scenario.run(runner) => First positional arg is the runner.
+        runner = args[0]
+        # -- RECORD test-runner state to restore it on each retry attempt.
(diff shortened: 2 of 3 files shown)


bittner:
Thanks! :1st_place_medal:

PR #1345, opened 2026-09-18
JUnitReporter: Create the report directory race-free

wiebren:
## Problem

`JUnitReporter.feature()` creates the report directory with a check-then-act sequence:

```python
if not os.path.exists(self.config.junit_directory):
    os.makedirs(self.config.junit_directory)
```

If several behave processes write their reports to the same `--junit-directory` (for example: a test run that is split up over multiple processes, in a CI job or with a parallel test runner), another process can create the directory between the check and `os.makedirs()`. The reporter then fails with a `FileExistsError` and the report of this feature is lost.

## Fix

Use `os.makedirs(..., exist_ok=True)`. The behaviour is otherwise unchanged (multiple directory levels are still created at once).

## Tests

Two unit tests in `tests/unit/reporter/test_junit.py`:

* nested report directories are created
* a directory that appears between the existence check and its creation is tolerated (the race is simulated by patching `os.path.exists`); this test fails with `FileExistsError` without the fix

Locally (Python 3.13, macOS): `pytest tests`, `behave features/` and `behave issue.features/` pass.

I will add the `CHANGES.rst` entry with the pull number in a follow-up commit.

---

Generated with Claude Code


behave/reporter/junit.py
```diff
@@ -290,9 +290,9 @@ def feature(self, feature):
         if self.show_hostname:
             suite.set('hostname', _text(gethostname()))
 
-        if not os.path.exists(self.config.junit_directory):
-            # -- ENSURE: Create multiple directory levels at once.
-            os.makedirs(self.config.junit_directory)
+        # -- ENSURE: Create multiple directory levels at once.
+        # RACE-FREE: Another process may create this directory concurrently.
```
bittner: Please remove both comments. They are superfluous. The commit should explain this briefly, the code should be self-explanatory.

CHANGES.rst
```diff
@@ -23,6 +23,7 @@ FIXED:
 * Include changes from ``behave v1.3.1`` (#1255, #1239)
 * issue #1028: use unittest.mock instead of mock (submitted by: pgajdos)
 * pull #1297: Context._pop(): Use internal config reference to avoid cleanup-errors if "context.config" is masked (provided by: wiebren)
+* pull #1345: JUnitReporter: Create the report directory race-free if several processes use the same junit directory (provided by: wiebren)
```
bittner, line 27: Please squash the commits as the changes belong together.

bittner (changes requested):
Thanks!

This is a small, targeted contribution, easy to approve. Good job!

Please squash the commits and remove the co-authored by trailer. We don't want advertising for LLM corporations here. And I trust in you, the human being, having at least glanced over the changes.

CHANGES.rst
wiebren: Done

behave/reporter/junit.py
wiebren: Done

bittner (approved):

bittner:
Thanks! :1st_place_medal:

PR #1348, opened 2026-09-18
CapturedFormatter: Fix crashes if a step has captured output

wiebren:
## Problem

`behave -f captured` crashes as soon as a step has captured output -- a passing step that prints something, and every failing step (its error message is captured output, too):

```
  File "behave/formatter/captured.py", line 95, in print_statement
    statement_name = "{statement.step_type.title()} {statement.name}".format(
AttributeError: 'str' object has no attribute 'title()'. Did you mean: 'title'?
```

`str.format()` cannot call a method in a replacement field (the f-string in the `PREPARED` comment above it can). Behind this one is a second crash:

```
  File "behave/formatter/captured.py", line 135, in make_capture_report
    report = captured_view.make_report(template=template, part_template=part_template)
TypeError: ManyCaptured.make_report() got an unexpected keyword argument 'part_template'
```

## Fix

* `print_statement()`: compute `step_type.title()` outside of the format string (kept `str.format()`, like the rest of this module).
* `make_capture_report()`: drop the `part_template` parameter. `ManyCaptured.make_report()` already renders its parts with `make_simple_report()` (template: `"{output}"`), which is what `part_template = "{output}"` asked for.

## Tests

New `features/formatter.captured.feature` (this formatter had no tests): a failing step and passing steps with captured output. Both scenarios fail without this change.

Locally (Python 3.13, macOS): `pytest tests`, `behave features/` and `behave issue.features/` pass.

The same code is in `release/v1.3.x`. I found it while looking into #1346; it is not related otherwise. I will add the `CHANGES.rst` entry with the pull number in a follow-up commit.

---

Generated with Claude Code


bittner:
Thanks for the fix! :1st_place_medal: 

Note that I squashed the commits and removed the co-authored by trailer. We don't want LLM corporation advertising here. The contribution affects your own reputation, so I trust that you looked at the changes and didn't submit everything blindly.

Issue #1012, opened 2022-04-06
Behave parallel executions


jcardenaspe:
When gonna be implemented a parallel execution feature?

jenisys:
Mmh, this is an open source project.
Therefore, if you need something, contribute to it (especially if you are in need of a feature).

The current tip of this Git repository provides an extension-point that allows you to provide your own test-runner.
Therefore, the easiest way is that YOU PROVIDE a parallel test runner implementation to the `behave community` and publish it as package on https://pypi.org .

If you have not enough "pain" to work on this and provide on your own (and figure out all the problems related to it), you need to wait until I have time to continue the work on my prototype. I have a pretty clear picture how the design should look like, but it is currently not really usable for the public (problems w/ bad behaving steps, shutdown or abortion of the test run and deadlocks related to the `multiprocessing` module).

OTHERWISE:
There are some `behave-parallel` implementations on Github.
But they seem to be outdated, meaning some Python2 / Python3 code cleanup was not performed yet (like with `python-modernise`). This means you cannot directly use it if you use Python3.

SEE ALSO:

* https://github.com/tokunbo/behave-parallel
* https://github.com/hugeinc/behave-parallel (forked)

SIMILAR:
* The latest `behave-parallel` version on https://pypi.org is not installable (`setup.py` seems to be missing)

russellsch:
Also worth a look: https://github.com/hrcorval/behavex

wiebren:
Also made a experimental addon package https://github.com/wiebren/behave-parallel-runner

PR #1349, opened 2026-09-19
Configuration: Remember how it was built (for parallel test runners); clarify --jobs

wiebren:
## Summary

Two small changes that align the core with parallel test runners, which live outside of the core as separately installable community runners (see #1012, #235):

1. `Configuration` remembers how it was built (`command_args`, `command_kwargs`, `command_load_config`), so that a test runner can build an equivalent configuration object in a worker process.
2. The `--jobs` option (and the runners appendix) says who supports it: not the default test runner, but separately installable community runners.

No behaviour of behave changes. The default runner still ignores `--jobs`.

## Problem

A parallel runner needs the configuration in its worker processes. A `Configuration` object cannot be sent to another process: it holds open output streams, reporters, a compiled tag-expression and arbitrary userdata values. So a worker has to build its own one, the same way the parent did:

```python
Configuration(command_args, load_config=load_config, **kwargs)
```

But `Configuration.__init__()` uses these three inputs and forgets them. A runner only receives the finished object, so it cannot know how it was built. Both parallel runners that use the runner extension point have to work around this, each in its own way:

* [behave-parallel-runner](https://github.com/wiebren/behave-parallel-runner) (mine) lets each worker re-run `Configuration()` and relies on `sys.argv` for the command args. That is correct for `behave ...` on the command line, but with a programmatically built configuration the user must describe by hand how it was built (`config.command_args = ...`, [README: Programmatic use](https://github.com/wiebren/behave-parallel-runner/blob/0e041eae52b1440846e404e9f7aa92003b19ba2f/README.md#programmatic-use)). On top of that, a whitelist of four values (`stage`, `lang`, `tags`, `userdata`) is [copied to the workers](https://github.com/wiebren/behave-parallel-runner/blob/0e041eae52b1440846e404e9f7aa92003b19ba2f/src/behave_parallel_runner/runner.py#L694-L743).
* [behave-pool](https://github.com/MathiasPaulenko/behave-pool) copies a fixed set of fields into a [`ConfigSnapshot`](https://github.com/MathiasPaulenko/behave-pool/blob/e784289e0cee95c6aca4ba9bc2397fd62a425f68/behave_pool/config.py#L92-L138) and [rebuilds the configuration from the paths only](https://github.com/MathiasPaulenko/behave-pool/blob/e784289e0cee95c6aca4ba9bc2397fd62a425f68/behave_pool/worker.py#L352-L371) (`load_config=False`). Whatever is not in the snapshot does not reach the workers, for example `--tags`, `-D name=value` and the config-file settings.

Neither workaround can be complete, because the information is gone by the time the runner gets the configuration object. Any future runner with worker processes runs into the same wall.

## Change

`Configuration` keeps its three inputs:

```python
command_args = self.make_command_args(command_args, verbose=verbose)
self.command_args = list(command_args)
self.command_kwargs = dict(kwargs)
self.command_load_config = load_config
```

`command_args` is stored *after* `make_command_args()`, so it is the resolved list: a `str` is split, and the `sys.argv[1:]` fallback is only in it if the program really is `behave`. A worker that is started from another program (a script, pytest) must not pick up that program's `sys.argv`.

A runner can then do:

```python
config2 = Configuration(config.command_args,
                        load_config=config.command_load_config,
                        **config.command_kwargs)
```

The attributes are declared in `Configuration.init()` like the other ones and are described in the class docstring (shown in the API docs).

### How to reproduce

```python
# -- FILE: run_programmatic.py
from behave.configuration import Configuration
from behave.__main__ import run_behave

if __name__ == "__main__":
    args = ["--jobs=2", "-r", "behave_parallel_runner:ParallelRunner",
            "--name=A1", "-f", "progress", "features"]
    config = Configuration(args, load_config=False)
    raise SystemExit(run_behave(config))
```

With four scenarios (`A1`, `A2`, `B1`, `B2`; `A2` and `B2` fail), `behave --name=A1` runs one of them:

| | Result |
|---|---|
| default runner, sequential | `1 scenario passed, 0 failed, 3 skipped` |
| parallel runner, behave v1.3.3 | `2 scenarios passed, 2 failed, 0 skipped` (workers never saw `--name=A1`) |
| parallel runner, this branch | `1 scenario passed, 0 failed, 3 skipped` |

behave-parallel-runner already reads exactly these attribute names, so it needs no change to benefit; its "Programmatic use" section becomes unnecessary.

### What this does not solve

Changes that are made to a configuration object *after* it was built (for example in a hook) are still not seen by a rebuilt one. That is up to the runner; making `Configuration` picklable would be a much bigger change and is not proposed here.

## `--jobs` option

The option exists in the core, but nothing in the core uses it. Its help text said "Only supported by test runners that support parallel execution", which leaves a user wondering where such a runner is. New text:

```
-j, --jobs, --parallel NUMBER
        Number of concurrent jobs to use (default: 1). Only
        supported by test runners that support parallel
        execution. The default test runner runs sequentially;
        parallel test runners are separately installable
        community runners (select one with --runner).
```

`docs/behave.rst` is regenerated with `docs/update_behave_rst.py` (only the two `jobs` entries change). The runners appendix gets a short section "Parallel Test Runners" with the same statement and the hint for implementors shown above. I did not name any runner package in the docs; that is your call (the ecosystem appendix would be the place).

## Testing

* `tests/unit/test_configuration.py`: new `TestConfigurationCommandArgs` (list, copy, `str`, `sys.argv` of `behave`, `sys.argv` of another program, kwargs, `load_config` default, and that an equivalent configuration can be built).
* `pytest tests`: 1648 passed, 3 skipped, 2 xfailed.
* `behave features/runner.*.feature features/configuration.*.feature features/cmdline.*.feature`: 198 scenarios passed.
* Sphinx build of `docs/`: no warnings (same as `main`).

`CHANGES.rst` has two entries; I will add the pull number to them once it is known.

---

Generated with Claude Code


bittner:
Thanks for the contribution! :1st_place_medal: 

Note that I removed the co-authored by trailers. We don't want advertising for LLM corporations here.

wiebren:
> Thanks for the contribution! 🥇
> 
> Note that I removed the co-authored by trailers. We don't want advertising for LLM corporations here.

Thank you, co author removal is fine by me, I just want to be open on the involvement from my end.

